PR #5, opened 2017-05-29
Use pointer_traits.to_address instead of addressof(*p)

glenfe:
`addressof(*p)` before calling `construct()` is technically undefined as there is no object yet constructed in the storage aliased by `p` (given that `addressof()` takes an l-value reference).

### DIFF
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -14,6 +14,7 @@
 #endif
 
 #include <boost/assert.hpp>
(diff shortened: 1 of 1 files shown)


danieljames:
There are a lot of other places where that would apply. I'll look into it, but I might just handle it directly.

glenfe:
Sounds good. I merged pointer_traits to master; you might want to also have Travis pull either develop or master  versions of repositories instead (or in addition to) the latest Boost release archive.

PR #8, opened 2018-07-09
Change type in enable_if parameters

sam-lunt:
When compiling with `-Wzero-as-null-pointer-constant`, parameters such as `typename boost::enable_if_c<true, void*>::type = 0` generate a warning. By changing `void*` to `int`, no warning should be generated.

### DIFF
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -353,7 +353,7 @@ namespace boost {
 
       template <class I>
       inline std::size_t insert_size(I i, I j,
-        typename boost::unordered::detail::enable_if_forward<I, void*>::type =
+        typename boost::unordered::detail::enable_if_forward<I, int>::type =
           0)
       {
         return static_cast<std::size_t>(std::distance(i, j));
@@ -361,7 +361,7 @@ namespace boost {
 
       template <class I>
       inline std::size_t insert_size(I, I,
-        typename boost::unordered::detail::disable_if_forward<I, void*>::type =
+        typename boost::unordered::detail::disable_if_forward<I, int>::type =
           0)
       {
         return 1;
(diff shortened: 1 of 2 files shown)


LeonineKing1199:
@sam-lunt Are you still interested in getting this PR merged? I've noted that it's quite old and has fallen behind develop quite a bit.

If you can, please rebase and we'll get the updated CI to run.

sam-lunt:
@LeonineKing1199 I rebased onto the newest version of develop. I think the changes are still valid as-is.

LeonineKing1199:
Hmm, seems like the GitHub Actions weren't triggered.

I think I've run into this issue in the past, actually. If you close and re-open the PR, the GHA tests should start running.

sam-lunt:
@LeonineKing1199 looks like the checks ran (and passed). I think it was waiting for approval from a project maintainer, so assuming that was you, thanks!

LeonineKing1199:
Alright, so I kicked this PR around to a few of the Boost authors and the direction we'd rather go is instead of changing the type in the `enable_if`s as they are now to silence the warning, we should be putting this `enable_if` in the return type.

If you're up for it, feel free to make the refactors. Otherwise, I'll just go ahead and update the code myself. Not a big deal either way.

sam-lunt:
My personal preference is actually to use a dummy template parameter. I think it's the most readable, since it doesn't obscure the function signature, it keeps the template parameter closer to the template constraints, and it can work with the default value for the second arg to `enable_if`:
``` cpp
template <class I, class = typename boost::unordered::detail::enable_if_forward<I>::type>
inline std::size_t insert_size(I i, I j)
```

That said, it's of course your prerogative to use whichever option you prefer. They'll be no hard feelings on my end : ) but I think it'll probably be faster if you make the update yourself, I don't think I'll have the time for a bit.

LeonineKing1199:
> That said, it's of course your prerogative to use whichever option you prefer. They'll be no hard feelings on my end : ) but I think it'll probably be faster if you make the update yourself, I don't think I'll have the time for a bit.

Heh, I feel that. I have a 9 month old at home and yeah, it's pretty much impossible to find free time (and more importantly energy) for anything :P

sam-lunt:
Ha, in that case, I know _exactly_ how you feel: just replace "9 month" with "5 month" : )

cmazakas:
Seems like it's been long enough. I'm going to be closing this.

PR #10, opened 2018-12-03
Using allocator's size_type to control sizeof unordered containers so that it is fixed in a 32/64 shd mem context

CerosDev:
Modified type of member variable, will now use the allocator traits's `size_type`.
This is useful when used in a 32/64bits Boost interprocess shared memory context. Using the previous `std::size_t` makes the container size different in both architecture, therefore impossible to store in a shared memory.

In most case, this makes no difference because the default allocator's `size_type` is `std::size_t`


LeonineKing1199:
@CerosDev Would you mind rebasing your branch against develop and then pushing so that we can get the CI running again?

CerosDev:
I opened this 3 years ago. I will just close this.

PR #23, opened 2021-05-05
pruned source-location from 32-bit C++03 build

tobias-loew:
32-bit C++03 build produces a lot of warnings whih led to a too big log output and the test failed


tobias-loew:
moved to other pr

PR #24, opened 2021-05-05
fix travis build

tobias-loew:
pruned source-location from 32-bit C++03 build
32-bit C++03 build produces a lot of warnings which led to a too big log output and the test failed
added python 3


tobias-loew:
@LeonineKing1199 thanks for the notification. I revoke this PR. Please have a look at #27, where I adopted drone for boost.unordered

Issue #22, opened 2021-03-20
scoped_allocator with unordered_map on macos


kleunen:
I have a user-defined type which is allocator aware:
https://github.com/kleunen/tilemaker/blob/win_ci_build/include/osm_store.h#L108-L122

This user-defined type I would like to store in an unordered_map, which is allocated in an mmap file using boost interprocess:
https://github.com/kleunen/tilemaker/blob/win_ci_build/include/osm_store.h#L331-L337

Finally I use emplace to store an entry in this map:
https://github.com/kleunen/tilemaker/blob/win_ci_build/include/osm_store.h#L375-L380

On MSVC and linux/g++ this works fine. However, if I compile on MacOS, with g++ and probably libc++ as STL, the unordered_map construct seems to default initialize the VarIntVector and not pass the scoped allocator. Which fails, because the interprocess allocator is not default constructable, it needs a reference to a segment manager. 

````
2021-03-20T07:14:56.5388750Z /usr/local/share/vcpkg/installed/x64-osx/include/boost/container/scoped_allocator.hpp:443:32: error: constructor for 'boost::container::dtl::scoped_allocator_adaptor_base<boost::interprocess::node_allocator<unsigned char, boost::interprocess::segment_manager<char, boost::interprocess::rbtree_best_fit<boost::interprocess::mutex_family, boost::interprocess::offset_ptr<void, long, unsigned long, 0>, 0>, iset_index>, 64>>' must explicitly initialize the base class 'boost::interprocess::node_allocator<unsigned char, boost::interprocess::segment_manager<char, boost::interprocess::rbtree_best_fit<boost::interprocess::mutex_family, boost::interprocess::offset_ptr<void, long, unsigned long, 0>, 0>, iset_index>, 64>' which does not have a default constructor
2021-03-20T07:14:56.5391710Z    BOOST_CONTAINER_FORCEINLINE scoped_allocator_adaptor_base()
2021-03-20T07:14:56.5392270Z                                ^
[...]

PR #31, opened 2021-09-21
Add explicit tests around use of `scoped_allocator_adaptor` to emulat…

LeonineKing1199:
…e issue https://github.com/boostorg/unordered/issues/22

### DIFF
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -8,6 +8,7 @@ import testing ;
 project unordered-test/unordered
     : requirements
         <warnings>all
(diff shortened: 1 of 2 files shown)


vinniefalco:
What's going on with the build failures?

LeonineKing1199:
> What's going on with the build failures?

Good news, I'm able to replicate the test failure locally. It only manifested in msvc 16.11.3.

I was using 16.11.1 and for some reason couldn't get it to fail so I updated my Visual Studio install and lo and behold, I have a local failure.

PR #33, opened 2021-11-17
Fix deprecated copy warnings and misc. C++20 `operator==` warnings

LeonineKing1199:


LeonineKing1199:
@pdimov I'm gonna close this PR and refactor it into 3 separate ones.

Namely, this PR introduced a broken copy assignment operator for the `test::cxx11_allocator` but this wasn't caught because of a defect in the memory tracker's reference counting.

I'm going to make a separate issue for each warning which will have its own PR and this way, everything should hopefully be a bit more trackable.

PR #40, opened 2021-11-18
Update SFINAE expressions to be in the return type

### LINKED ISSUE #39: `-Wzero-as-null-pointer-constant` generates warnings in SFINAE
Users have noted that certain warnings will trigger for the SFINAE used in the main `implementation.hpp` file as noted here https://github.com/boostorg/unordered/pull/8

Due to the busy lives of contributors, this issue can be directly tackled by a separate PR which refactors the SFINAE into the return type thus fixing the error and also updating the code to be more in line with current style recommendations.

LeonineKing1199:
Closes #39 and overrides https://github.com/boostorg/unordered/pull/8

### DIFF
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -352,17 +352,16 @@ namespace boost {
       // insert_size/initial_size
 
       template <class I>
-      inline std::size_t insert_size(I i, I j,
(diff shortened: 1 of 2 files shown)


glenfe:
Looks like a healthy mix of `boost::enable_if` and `boost::enable_if_c`.  I'd probably pick one (and that would be `boost::enable_if_` from TT).

pdimov:
Good enough for government use.

PR #45, opened 2021-11-22
Transparent count unordered map

### LINKED ISSUE #44: `unordered_map::count()` is missing transparent overload
[P1690R1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1690r1.html) added a transparent overload of `count()` that avoids creating an intermediate object when getting the count.

See: https://en.cppreference.com/w/cpp/container/unordered_map/count

LeonineKing1199:
Closes #44

### DIFF
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -726,16 +726,15 @@ namespace boost {
 ////////////////////////////////////////////////////////////////////////////
 // Type checkers used for the transparent member functions added by C++20 and up
 
-      template <class>
-      struct make_dependent : public true_type {};
-
-      template <class, class = void> struct is_transparent : public false_type
+      template <class, class, class = void>
+      struct is_transparent : public false_type
       {
       };
 
-      template <class T>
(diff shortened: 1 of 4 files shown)


LeonineKing1199:
@pdimov @glenfe 

I think this PR is largely good to go. I'm not 100% sure on if this test suite is exhaustive enough so if you can think of anything to add, I'd greatly appreciate it.

LeonineKing1199:
@pdimov 

The test has been cleaned up so that the `key`'s assignment operator no longer exists as it's not needed. We've also refactored 3/4 of the test cases into their own function and there's now a helper function to clear all the `static` counts we use.

pdimov:
This is fine, but you might want to move the transparent test into its own function as well; both for symmetry and because you'll probably need to do that anyway when you add `unordered_multimap` to the test.

LeonineKing1199:
> This is fine, but you might want to move the transparent test into its own function as well; both for symmetry and because you'll probably need to do that anyway when you add `unordered_multimap` to the test.

Noted and updated!

PR #53, opened 2021-12-01
Implement heterogeneous `equal_range()` for `unordered_map`

### LINKED ISSUE #49: `unordered_map::equal_range` needs transparent overload
Described [here](https://en.cppreference.com/w/cpp/container/unordered_map/equal_range).

LeonineKing1199:
Closes #49


pdimov:
The test is incomplete. The map should have more than one element, and the test should check whether the returned range is of the correct length and whether the key is as expected (not just whether the value is 1337.)

PR #54, opened 2021-12-03
Heterogeneous erase

### LINKED ISSUE #46: `unordered_map::erase` needs transparent overload
Outlined [here by cppref](https://en.cppreference.com/w/cpp/container/unordered_map/erase).

LeonineKing1199:
Closes #46


LeonineKing1199:
@pdimov This PR implements heterogeneous `erase()`.

Note, this one took a little more refactoring than the others.

We needed ways of probing the forward list for the `prev` node in order to do the erasing like it's done now so I added an `_impl` variant that is generic over the binary predicate and the key type to test against.

I didn't remove `erase_key_unique()` yet because it's depended on by `unordered_set` too.

PR #55, opened 2021-12-06
Heterogeneous Extract

### LINKED ISSUE #47: `unordered_map::extract` needs transparent overload
More info [found here](https://en.cppreference.com/w/cpp/container/unordered_map/extract).

LeonineKing1199:
Closes #47


pdimov:
There should be a separate `convertible_to_const_iterator`, instead of changing the current one, and both should probably be tested.

PR #56, opened 2021-12-07
Replace internal implementation of `allocation_traits` with Core's

### LINKED ISSUE #30: Internal `allocator_traits` should be replaced by `core/allocator_access`
Unordered could use some sprucing up by replacing its own homespun allocator traits with the utilities in Core:
https://www.boost.org/doc/libs/develop/libs/core/doc/html/core/allocator_access.html

LeonineKing1199:
Closes #30


glenfe:
Get rid of rebind_wrap. Why is it needed? There can't be too many places using it.

glenfe:
Also, why even support BOOST_UNORDERED_USE_ALLOCATOR_TRAITS at all? i.e. Always unconditionally just use <boost/core/allocator_traits.hpp>

pdimov:
> i.e. Always unconditionally just use <boost/core/allocator_traits.hpp>

That's what it does?

> Get rid of rebind_wrap.

Yeah, in a subsequent PR, let's merge this first.

The Container include should be removed.

glenfe:
I noticed the `BOOST_UNORDERED_USE_ALLOCATOR_TRAITS ` macro still present, but now I gather from the last comment that its removal will similarly be in a follow-up PR.

pdimov:
It's checked in the tests so the removal should be a separate diff.

LeonineKing1199:
@glenfe How would you like the failing test case (the inherited SOCCC one) to be handled? For example, is it acceptable to simply `#ifdef` this test out for C++03 mode?

glenfe:
It's worth it to have a test that checks SOCCC otherwise I did the work for nothing. :) Have an 03 test with an allocator that does define a SOCCC member that is detected with our current detection emulation.

pdimov:
There are several such tests, and only one of them is failing. These pass:
https://github.com/boostorg/unordered/blob/00b504ebc51e1ecda9347106ed0c91dba77c651a/test/unordered/allocator_traits.cpp#L130
https://github.com/boostorg/unordered/blob/00b504ebc51e1ecda9347106ed0c91dba77c651a/test/unordered/allocator_traits.cpp#L172
https://github.com/boostorg/unordered/blob/00b504ebc51e1ecda9347106ed0c91dba77c651a/test/unordered/allocator_traits.cpp#L278

This one fails, because the member function is defined in a base class:
https://github.com/boostorg/unordered/blob/00b504ebc51e1ecda9347106ed0c91dba77c651a/test/unordered/allocator_traits.cpp#L216

glenfe:
In that case, yeah, only enable that one in when `!defined(BOOST_NO_CXX11_ALLOCATOR)`.

PR #60, opened 2021-12-15
Multimap heterogeneous `find()`

cmazakas:


include/boost/unordered/detail/implementation.hpp
```diff
@@ -713,6 +713,11 @@ namespace boost {
       {
       };
 
+      template <class X, class H, class P>
+      struct is_transparent2 {
+        static bool const value = is_transparent<X, H>::value && is_transparent<X, P>::value;
```
glenfe, line 718: If `is_transparent` is only used in `is_transparent2` it does not need the first template parameter anymore.

```
template<class, class = void>
struct is_transparent
    : false_type { };
    
template<class T>
struct is_transparent<T, typename make_void<typename T::is_transparent>::type>
  : true_type { };

template<class, class A, class B>
struct are_transparent
    : conjunction<is_transparent<A>, is_transparent<B> > { };
```

PR #67, opened 2021-12-27
Set Heterogeneous `equal_range()`

cmazakas:


pdimov:
Looks good. We should however test `equal_range(1)` as well, so that the multiset case tests both a range of size 1 and a range of size >1.

PR #76, opened 2022-01-07
Fix Change Log TOC

cmazakas:
By updating the Jamfile, we can get the Unordered change log to render like the one in Core.

### DIFF
--- a/doc/Jamfile.v2
+++ b/doc/Jamfile.v2
@@ -14,7 +14,7 @@ xml unordered : unordered.qbk :
 
 boostbook standalone : unordered :
     <xsl:param>chunk.first.sections=1
-    <xsl:param>chunk.section.depth=2
+    <xsl:param>chunk.section.depth=1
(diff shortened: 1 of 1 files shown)


glenfe:
I would probably have kept the changelog using [heading] instead of [section] and leave the TOC depth as it is now.

pdimov:
Yep. Let's just get back to `[h2]` as it was before [the change](https://github.com/boostorg/unordered/commit/8ce147dcbdddc520c77f327a410eaa8b234e2a64).

glenfe:
[heading] might be better than [h2] because it automatically chooses the heading size based on the current section depth.

pdimov:
Since the Unordered docs are of the "integrated" variety, the local chunking parameter doesn't actually apply. They currently [look like this](https://www.boost.org/doc/libs/develop/doc/html/unordered.html). I think that we don't want this, so using `[heading]` seems the way to go.

PR #78, opened 2022-01-11
Fix integer overflow UB in test suite

cmazakas:
Enable the CI to run with `address-sanitizer=norecover undefined-sanitizer=norecover`.


pdimov:
`static_cast<unsigned>(x)` please.

PR #82, opened 2022-01-13
Implement `unordered_multiset::contains()`

cmazakas:


cmazakas:
`contains()` is indeed missing. That'll will come in the next PR.

As far as I'm aware, everything else that needs to be documented is documented but I'll look over the docs anyway just to make sure.

PR #105, opened 2022-02-23
Implement `erase_if`

cmazakas:


cmazakas:
@pdimov This one should be good to go; I was curious if the wording around the function is sufficiently clear. Let me know if you think it needs some work.

PR #109, opened 2022-03-01
`unordered_map` reference doc fixes

### LINKED ISSUE #101: Missing reference docs for `try_emplace()`
`unordered_map` is missing documentation for the `try_emplace` series of member functions

cmazakas:
Closes #101 and #102 

We were missing some documentation for implemented member functions specific to `unordered_map`. This PR aims to add that missing documentation, keeping it in line stylistically with what's existing.


cmazakas:
@pdimov This one should be good to go. I tried to keep these additions in line with how the reference docs previously did things which was have separate entries for the versions with a `hint` and those without because the return type defers.

PR #110, opened 2022-03-01
Add missing `std::` qualification to usages of `pair` in the reference

cmazakas:
Unfortunately, some mentions of `pair` in the reference docs were missing appropriate `std::`-qualification so this PR aims to address that.


pdimov:
Is this one ready as well?

cmazakas:
Sorry for the late reply, this one should be good to go as well.

PR #115, opened 2022-06-02
Add Boost::align dependency to CMake build

Flamefire:
This is required after #112

@cmazakas @pdimov Please merge soon as this currently breaks the super-project CMake build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ target_include_directories(boost_unordered INTERFACE include)
 
 target_link_libraries(boost_unordered
   INTERFACE
+    Boost::align
     Boost::assert
     Boost::config
     Boost::container


pdimov:
Not sure that the use of Align is intended.
```
align:
    <boost/align/alignment_of.hpp>
        from <boost/unordered/detail/fca.hpp>
```
alignment_of is in TypeTraits.

pdimov:
Yeah, the include is unnecessary as the code doesn't use it. It uses `boost::alignment_of` from TypeTraits. https://github.com/boostorg/unordered/blob/e36e3bcf96ef1d0655544d6ae6ef0a2fa14ccda5/include/boost/unordered/detail/fca.hpp#L409

Flamefire:
Oh, right. --> #116

PR #119, opened 2022-06-06
Fix `-Wmaybe-unitialized` warning in gcc-12

cmazakas:
Optimizations in gcc-12 seem to call out dubious usage of `reinterpret_cast` that hasn't been laundered. We add a c++03 compatible shim because of those same optimizations

### DIFF
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -1548,6 +1548,21 @@ namespace boost {
       // If an exception is thrown between these two calls, use
       // 'cleanup_spare_functions' to destroy the unused constructed functions.
 
+#if defined(_GLIBCXX_HAVE_BUILTIN_LAUNDER)
+      // gcc-12 warns when accessing the `current_functions` of our `functions`
+      // class below with `-Wmaybe-unitialized`. By laundering the pointer, we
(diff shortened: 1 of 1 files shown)


pdimov:
static_cast is "more correct" than reinterpret_cast, there's no need to change this for the worse.

BOOST_UNORDERED_LAUNDER does not adhere to the naming convention for such macros, should be BOOST_UNORDERED_HAS_LAUNDER; and since it's only used locally, should be undefined afterwards.

There's no need to check for C++17 and use std::launder in an `#ifdef _GLIBCXX_HAVE_BUILTIN_LAUNDER` block; if we're there we know that `__builtin_launder` is available, so just use that.

cmazakas:
Out of curiosity, why is `static_cast` better here? What's inappropriate about `reinterpret_cast` vs it?

Ah, I think I can figure out why...

Is it because we're not strictly reinterpreting something pointer interconvertible with the `function_pair`? I realize now, I likely broke the strict rules around reinterpret_cast'ing something because we're working with a simple POD type so this would be an aliasing violation.

pdimov:
The result of `reinterpret_cast` is implementation-defined in the general case, whereas static casting to/from void* is not. In this specific case it probably doesn't matter, but downgrading the static_cast sequence is bad practice; there was nothing wrong with it.

PR #121, opened 2022-06-08
Update Appveyor from Boost.CI 

Flamefire:
Extracted from #114 to make the changes easier to handle

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -54,18 +54,36 @@ environment:
     - FLAVOR: Visual Studio 2019
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
       B2_CXXFLAGS: -permissive-
-      B2_CXXSTD: 14,17,2a
+      B2_CXXSTD: 14,17
       B2_TOOLSET: msvc-14.2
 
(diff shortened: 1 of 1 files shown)


Flamefire:
@cmazakas This and #114 are ready now. I'd suggest to merge them both at the same time (at least on GHA the 2nd commit/merge will cancel the CI run started by the first) and then merge develop to master, potentially after adjusting the `.codecov.yml` to your liking, see #114 for details. Only then the codecov settings will be in effect.

pdimov:
@Flamefire In order to fix the Cygwin failure (marked as allowed now), we need to pass `include=` to `b2`, but at the moment boost-ci doesn't seem to have the `B2_INCLUDE` option: https://github.com/boostorg/boost-ci/blob/a5712f1268770e1a32e352d485247d0460145a3f/ci/build.bat#L13-L18

Flamefire:
IMO this needs to be set in the Jamfiles. I wouldn't expect users to need to know which paths to pass to `include=`.

pdimov:
I am the user in this case and I know which paths I need to pass to get Cygwin to work.

PR #124, opened 2022-06-10
Add Drone CI

Flamefire:
Supersedes and closes #27

I simply C&P the config file from Boost.CI, so not sure if everything passes.

Notes on Drone (if not already done, can you do that please @cmazakas ?):

> - To activate Drone, visit https://drone.cpp.al/. Authorize Drone: Click the "Authorize cppalliance-drone" button. Sync repositories: Click the "sync" button. A list of repositories will appear. For the relevant repo, click and then choose "Activate Repository". In the settings page, change Configuration from .drone.yml to .drone.star. "Save".
>
> More pointers about Drone:
>
> -    "asan" jobs require elevated privileges. Contact an administrator or open an issue at [drone-ci](https://github.com/CPPAlliance/drone-ci) to set your drone repository to "Trusted".
> -    If not using asan, simply remove the jobs.

CC @tobias-loew. Just want to make sure your work doesn't get forgotten.


Flamefire:
@cmazakas For Codecov on Drone you need to copy the "Repository Upload Token" from the settings page of the repo on [Codecov](https://codecov.io) and use it to create a new secret named `codecov_token` on the settings page of the repo on [Drone](https://drone.cpp.al).

Besides that it passes: https://drone.cpp.al/Flamefire/unordered/3 (now: https://drone.cpp.al/Flamefire/unordered/4)

And it seems the codecov bot is now running. It detects that there is some dead code in `test/objects/test.hpp`: The int-functor part of the `hash` class is only called on a default constructed `hash`-object.

pdimov:
What's the purpose of this? The benefit of Drone is that we can use it to cover configurations GHA lacks (GCC 4.4, 4.6, 4.7, ARM, s390x, etc), but this PR doesn't do this, it just duplicates the same configs already covered.

pdimov:
Here's my Drone variation: https://github.com/boostorg/system/blob/develop/.drone.jsonnet (self-contained as usual)

I prefer jsonnet, see if you like it better. We didn't have s390x yet when I wrote it, but it's an important configuration because it's big endian.

Flamefire:
Redundancy, as that was also a reason to add Drone.
I just used the config as it is on boost-ci. If there are more supported and useful then those should be added to boost-ci, which is the blueprint / testbed for the other repos to copy from. I don't think it is feasible to go looking in random boost repos for a good config.
So if yours is better, can you add/replace it on boost-ci with instructions how to use it?

pdimov:
drone.cpp.al has limited capacity and I don't think we should be carpet bombing it with jobs we don't actually need because they are already covered elsewhere. This will ruin it for everyone.

Flamefire:
Sure. Closing this then.

PR #128, opened 2022-06-20
Remove dependencies on Iterator, Detail

cmazakas:
Update test CML to include ConceptCheck as a dependency as it's no longer transitively included.

### DIFF
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -2449,13 +2449,8 @@ namespace boost {
             k, this->hash_function(), this->key_eq());
         }
 
-#if defined(BOOST_MSVC)
-#pragma warning(push)
-#pragma warning(                                                               \
-  disable : 4714) // function 'function' marked as __forceinline not inlined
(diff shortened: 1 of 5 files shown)


pdimov:
Use `conditional` from type_traits instead of declaring this MSVC 6 if-then thing inline maybe?

pdimov:
You shouldn't need the user-declared copy constructor and destructor of `grouped_bucket_iterator`. The compiler-generated ones should be fine.

pdimov:
Same for `grouped_local_bucket_iterator`, including the copy assignment.

pdimov:
Same for `const_grouped_local_bucket_iterator`.

pdimov:
Same for `iterator` and `c_iterator`.

include/boost/unordered/detail/implementation.hpp
```diff
@@ -1745,6 +1743,50 @@ namespace boost {
 
           iterator() : p(), itb(){};
```
pdimov, line 1742: Extra semicolon here.

Issue #139, opened 2022-08-16
unordered_set move leads to crash in 1.80.0


ergpudb:
This code segfaults in 1.80.0 but not 1.79.0 (nor using std::unordered_set):

```
    boost::unordered_set<std::string> a;
    boost::unordered_set<std::string> b = std::move(a);
    boost::unordered_set<std::string> c = std::move(a);
    c.clear(); // <-- segfault
```

Based on the standard, the first move should leave `a` in a "valid but unspecified state"; since `a` should still be valid, moving from it a  second time presumably should not corrupt `c` (even if the contents of `c` are technically unspecified afterward).

Note: it appears that unordered_map also exhibits the same behavior.

ergpudb:
Even simpler (no double move):

```
    boost::unordered_set<std::string> a;
    boost::unordered_set<std::string> b = std::move(a);
    a.clear(); // <-- segfault
```

cmazakas:
Thank you for the bug report!

I've replicated this issue locally and am aware of the cause.

I'm working on the fix.

cmazakas:
@ergpudb this fix will be landing in 1.81. In the interim, I'm going to close the issue.

PR #140, opened 2022-08-17
Fix use-after-move segfault

cmazakas:
Addresses https://github.com/boostorg/unordered/issues/139

Add `post_move_tests` which test the moved-from state of all the unordered containers, including move construction, move assignment, move with equal/unequal allocators and POCMA.


include/boost/unordered/unordered_map.hpp
```diff
@@ -2069,6 +2069,10 @@ namespace boost {
     template <class K, class T, class H, class P, class A>
     float unordered_map<K, T, H, P, A>::load_factor() const BOOST_NOEXCEPT
     {
+      if (table_.size_ == 0) {
+        return 0.0;
```
pdimov, line 2073: Should be `0.0f`.

PR #141, opened 2022-08-19
Split 64-bit mingw appveyor job into two to help prevent timeouts

cmazakas:


pdimov:
As a general rule later standards are slower to compile so maybe the split should be 03,11,14 vs 17,20. But wait for it to finish first.

PR #143, opened 2022-08-24
Update internal table to no longer allocate on default constructions and when the bucket count is 0

cmazakas:
Also make sure assignment operators don't allocate either as:
```cpp
T x;
T y;
x = y;
```
heavily implies no allocations should be happening.

### DIFF
--- a/include/boost/unordered/detail/fca.hpp
+++ b/include/boost/unordered/detail/fca.hpp
@@ -494,11 +494,25 @@ namespace boost {
         group_pointer groups;
 
       public:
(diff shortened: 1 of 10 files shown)


joaquintides:
The following:
```cpp
boost::unordered_set<int> s;
assert(s.bucket_count()==0);
s.rehash(0);
std::cout<<s.bucket_count()<<"\n";
```
prints 13, bucket count should be kept to 0 in this case, IMHO. Same thing for `reserve(0)`.

pdimov (approved):

PR #144, opened 2022-09-02
Rehashing Conformity

cmazakas:
Update the rehash() and reserve() functions so that they behave similarly to libstdc++ and libc++, i.e. rehashing should grow and shrink the allocation accordingly while still respecting the max load factor

### DIFF
--- a/include/boost/unordered/detail/fca.hpp
+++ b/include/boost/unordered/detail/fca.hpp
@@ -494,6 +494,14 @@ namespace boost {
         group_pointer groups;
 
       public:
+        static std::size_t bucket_count_for(std::size_t num_buckets)
+        {
+          if (num_buckets == 0) {
+            return 0;
+          }
+          return size_policy::size(size_policy::size_index(num_buckets));
+        }
+
         grouped_bucket_array()
             : empty_value<node_allocator_type>(
                 empty_init_t(), node_allocator_type()),
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -2524,6 +2524,17 @@ namespace boost {
           new_buckets.insert_node(itnewb, p);
         }
 
+        static std::size_t min_buckets(std::size_t num_elements, float mlf)
+        {
+          std::size_t num_buckets = static_cast<std::size_t>(
+            std::ceil(static_cast<float>(num_elements) / mlf));
+
+          if (num_buckets == 0 && num_elements > 0) { // mlf == inf
+            num_buckets = 1;
+          }
+          return num_buckets;
+        }
+
         void rehash(std::size_t);
         void reserve(std::size_t);
         void reserve_for_insert(std::size_t);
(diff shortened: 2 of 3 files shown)


joaquintides (commented):
* Consider using `std::floor` instead of `floor`.
* Why are you using `floor` in `recalculate_max_load` anyway? The `static_cast<std::size_t>(f)` bit already does the truncation, right? 
* You're using `std::size_t` sometimes and `size_type` some other times.
* `implementation.hpp#L3454`: you're introducing `size_type bc = num_buckets` when you could have continued using `num_buckets` (which is no longer referred to).
* `prime_fmod_size` is only used by rehash in the context of `table`. This is breaking encapsulation, as the growing policy is an implementation detail of `grouped_bucket_array`. Consider adding a static memfun like `bucket_count_for(num_buckets)` in `grouped_bucket_array`. This memfun should also cover the case `num_buckets`=0 (returning 0).
* The calculation of `min_required` is incorrect due to the added 1.0, as `x.rehash(x.bucket_count())`, which must do nothing by definition, will *grow* the number of buckets when the size is exactly the maximum load. I understand this is here to cover the case `mlf_`=∞, but I'm afraid you need to reformulate. Please add a test case before fixing.
* I recommend trying to rewrite `rehash` so that you don't have a `size_==0` branch. The code could look something like this:

```cpp
std::size_t min_buckets(std::size_t num_elements,float mlf)
{
  std::size_t m=static_cast<size_type>(std::ceil(static_cast<float>(num_elements)/mlf));
  if(num_elements&&!m)m=1; // mlf==inf
  return m;
}

template <typename Types>
inline void table<Types>::rehash(std::size_t num_buckets)
{
  num_buckets=(std::max)(
    buckets_.bucket_count_for(num_buckets),
    min_buckets(size_,mlf_));
  if(num_buckets!=this->bucket_count())this->rehash_impl(num_buckets);
}
```

* `implementation.hpp#L3475`: why the `size_ == 0` in the conditional? If `num_elements > 0 && num_buckets == 0`, that implies that `mlf_`=∞, in which case 1 is the right choice for the requested `num_buckets` regardless of whether there are elements in the container or not.
* The standard [says](http://eel.is/c++draft/unord.req#general-236) that `reserve` is exactly `rehash(ceil(num_elements / max_load_factor()))`, which is `rehash(0)` when `mlf_`=∞. We're not following this requirement as we're issuing `rehash(1)` here, which is IMHO the correct behavior --if `num_elements`>0, I certainly want at least one bucket. As I see it, this is a defect in the standard and you may want to file a DR: either the definition of `reserve` should be changed or else ∞ be banned as a valid argument to `max_load_factor`.
* This new behavior is not documented, and it should as it goes beyond the standard requirements.

joaquintides:
Looks good to me. There's a lot of commented-out code in `rehash_tests.cpp`. After a private conversation with Chris, I understand that doc updates will arrive in a different PR --I'd rather have it all in one single PR, but this other approach works as well.

pdimov:
The GHA and Appveyor failures are transient, but the Drone gcc 4.7 one looks legitimate.

joaquintides:
Ouch, didn't see that one, thought they were all transient stuff.

PR #145, opened 2022-09-12
Fix move assign noexcept violation/perf regression

cmazakas:
An unconditional `reserve()` call was added to move assignment breaking noexcept invariants and pessimisizing performance when POCMA is false but the allocators still compare equal.

This fixes a gcc-4.7 CI failure in #144 and should be merged first.

### DIFF
--- a/include/boost/unordered/detail/implementation.hpp
+++ b/include/boost/unordered/detail/implementation.hpp
@@ -2374,10 +2374,6 @@ namespace boost {
         template <typename UniqueType>
         void move_assign(table& x, UniqueType is_unique, false_type)
         {
-          if (x.size_ > 0) {
-            reserve(x.size_);
-          }
-
           if (node_alloc() == x.node_alloc()) {
             move_assign_equal_alloc(x);
           } else {
--- a/test/unordered/noexcept_tests.cpp
+++ b/test/unordered/noexcept_tests.cpp
@@ -11,6 +11,7 @@
 // clang-format on
 
 #include "../helpers/test.hpp"
+#include "../helpers/fwd.hpp"
 
 #if defined(BOOST_MSVC)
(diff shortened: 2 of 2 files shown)


pdimov (approved):

joaquintides (approved):
OK by me. Can you explain to me why this spurious `reserve` triggered a *termination* on GCC4.7 and nowhere else?

cmazakas:
> OK by me. Can you explain to me why this spurious `reserve` triggered a _termination_ on GCC4.7 and nowhere else?

Yup!

gcc-4.7 is a fringe compiler in that it's C++11 support isn't 100%. So we were in a case where the compiler supports `noexcept` and has `std::allocator::is_always_equal` being `true_type`. This means the noexcept specification of `unordered_set::operator=(unordered&&)` can be true. However, POCMA was false.

In the original noexcept_tests, we were move-assigning a small unordered_set into a bigger unordered_set so the erroneous call to reserve() had no effect as reserve() didn't used to shrink the number of buckets.

The first thing I did for this PR was add a test case doing the reverse: assigning a big unordered_set into a small one to trigger a proper rehashing. Because we actually started rehashing, the Hasher was invoked which was set to throw and because noexcept was true, we got a terminate call.

The implementation was internally bifurcated around POCMA. The reserve() call was erroneously placed in the POCMA => false branch and wasn't guarded by an allocator equality check, i.e. only call `reserve()` when POCMA is false and allocators are unequal.

The test now checks that `a.allocate()` isn't called either.

joaquintides:
Ok, so GCC 4.7 is the *only* example where `std::allocator` has no POCMA typedef (and hence POCMA is deduced to be false) *and* `noexcept` is supported, right?

Thanks for th explanation. Merging now.

PR #153, opened 2022-10-29
Feature/foa anti drift

joaquintides:
Consider this scenario of repetated insert/erase runs:
```cpp
boost::unordered_flat_set<int>     s;
std::uniform_int_distribution<int> dist;
std::mt19937                       gen(0);

for(int j=0;j<10;++j){
  // insert
  gen.seed(j);
  for(int i=0;i<1720000;++i)s.insert(dist(gen));

  // do some lookup

  // erase
  gen.seed(j);
  for(int i=0;i<1720000;++i)s.erase(dist(gen));
}
```
After each insertion run, the average number of hops on unsuccessful lookup is:
```
0.233472
0.490452
0.768257
1.07312
1.41655
1.80602
2.24332
2.73784
3.32208
3.98326
```
that is, probe lengths grow unboundedly (numbers obtained through instrumentation). This _drift_ phenomenon is due to the fact that overflow bytes can't be reset as elements are erased, so they keep saturating on new insert runs --all non-relocating, open-addressing containers, like Abseil for example, experience this problem.
The anti-drift mechanism proposed consists in reducing the maximum allowed load when erasing an element that potentially caused an overflow, so that, on the subsequent insert run, rehashing is triggered _before_ hitting the load of the previous run --on average, when the load reaches (previous load − number of erased elements × 0.18). I've re-run the experiment above with anti-drift and the average number of hops stays at 0.23 between runs, as it should.  In this context, rehashing does not grow the bucket array (the same capacity as before is allocated), though we introduce a _hysteresis factor_ to actually grow the array if insert/erase runs oscillate minimally in the vicinity of full load.

Performancewise, anti-drift only affects erasure. [These benchmarks](https://github.com/joaquintides/boost_unordered_benchmarks/tree/foa_anti_drift) show that the overhead ranges from negligible to ~10% in non-SIMD (see "Running erasure").

As for the container's interface, the obvious impact is that rehashing can happen before hitting the theoretical maximum load. We can do two things here (**open question to you**): 
* Leave `load_factor()` as is and let users know the returned valor is a hint and rehashing can happen before expected if there were previous erasures.
* Change the definition of `load_factor()` to return `(size() + amount detracted to maximum load)/bucket_count()`.

### DIFF
--- a/include/boost/unordered/detail/foa.hpp
+++ b/include/boost/unordered/detail/foa.hpp
@@ -1660,11 +1660,11 @@ private:
      * when the number of erased elements is <= 10% of total elements at full
      * load, which is implemented by requesting additional F*size elements,
      * with F = P * 10% / (1 - P * 10%), where P is the probability of an
-     * element having caused overflow; P has been determined to be ~0.187 via
-     * simulation, yielding F ~ 0.019 ~ 1/52.
+     * element having caused overflow; P has been measured as ~0.162 under
+     * ideal conditions, yielding F ~ 0.0165 ~ 1/61.
      */
     auto     new_arrays_=new_arrays(std::size_t(
-               std::ceil(static_cast<float>(size_+size_/52+1)/mlf)));
+               std::ceil(static_cast<float>(size_+size_/61+1)/mlf)));
     iterator it;
     BOOST_TRY{
       /* strong exception guarantee -> try insertion before rehash */
--- a/include/boost/unordered/detail/foa.hpp
+++ b/include/boost/unordered/detail/foa.hpp
@@ -1327,7 +1327,7 @@ public:
   void erase(const_iterator pos)noexcept
   {
     destroy_element(pos.p);
-    reset_slot(pos.pc);
+    recover_slot(pos.pc);
   }
 
   template<typename Key>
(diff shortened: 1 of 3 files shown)


pdimov:
I assume you mean `max_load_factor()`, because `load_factor()` returns the current load factor (size / capacity). If so, I prefer option 2, make it return the real max load factor, beyond which reallocation happens.

Since this is a new container, we might consider actually exposing max_load(), which is a more useful query, and typically what users need (how many elements can I insert without reallocation?).

joaquintides:
> I assume you mean `max_load_factor()`, because `load_factor()` returns the current load factor (size / capacity). If so, I prefer option 2, make it return the real max load factor, beyond which reallocation happens.

I really meant `load_factor()`, rationale being that slots not recovered count as "load". As for changing the definition of `max_load_factor()`, I don't know if users expect this number to change given its static nature in closed addressing.

Anyway, I like your `max_load()` idea very much: it retains the standard meaning of `load_factor()` and `max_load_factor()` and gives a simple way for users to plan ahead of rehashing.

pdimov:
> I really meant `load_factor()`, rationale being that slots not recovered count as "load".

Oh I see. Yes, this makes sense. I still prefer option 2, then, from the two given.

joaquintides:
> Oh I see. Yes, this makes sense. I still prefer option 2, then, from the two given.

But we both prefer option 3, do we? (I do)

3. Do not touch `load_factor()` or `max_load_factor()`, and provide `max_load()`

pdimov:
Yes, I prefer option 3 too.

PR #155, opened 2022-10-31
Updated 1.81 documentation

joaquintides:
Hi,

This is, as far as I can see, a 100% complete update of Boost.Unordered docs for 1.81 including `boost::unordered_flat_set` and `boost::unordered_flat_map`. We've got plenty of time to polish this, so, I encourage you to download, build and thoroughly review --I know it's a drag, but users will thank us. In particular, I recommend that you try to read it from the beginning to the end so as to assess whether the new parts dovetail well with previous material.

Some observations and open questions:

* I've tried to keep the style simple and fairly entry-level, which is how the original documentation felt to me.
* Both sets of benchmarks (for fca and foa) are now kept in the same repo (https://github.com/joaquintides/boost_unordered_benchmarks) under different branches, and references to this from docs have been updated. Could it be a good idea to transfer this repo to boostorg? Now it's in my personal space.
* Docs say flat containers require C++11 (we've managed to make it work with rather old C++11 compilers), but we can still raise that bar docwise to C++14.
* I haven't added for foa an analogous section to that explaning fca structure because I don't think the latter belongs in the docs anyway, given we have more extensive articles we can point to. Your opinion?
* I've taken the liberty to add Peter and myself to the list of (C) authors at the end.

So, please take some quality time to review this and give fedback. Thank you!


pdimov:
> * Could it be a good idea to transfer this repo to boostorg? Now it's in my personal space.

We can do that. You'll need to transfer the repo to me, and I'll transfer it to boostorg. But it might be better not to, because if it's under boostorg GHA will get throttled.

pdimov:
> * I haven't added for foa an analogous section to that explaning fca structure because I don't think the latter belongs in the docs anyway, given we have more extensive articles we can point to. Your opinion?

Unfortunately, I think that the right thing to do here is to have such a section. Boost docs, traditionally, have been kept self-contained so as to enable offline reading from the release tarball, without requiring internet access.

joaquintides:
* Transfer of [boost_unordered_benchmarks](https://github.com/joaquintides/boost_unordered_benchmarks) to `boostorg`: GHA runs on Sam's dedicated servers, so I understand we won't have any problem with throttling?
* Section on foa data structure: OK, I'll write the full-fledged external article (around next week) and then we can paste here an abridged version.
* Your Slack comment about a typo in `operator==` reference ("behavior is undefined if the two containers don’t have equivalent equality predicates"). Re-reading it, I'm not sure what problem you're referring to. Is it about mentioning that `Key::operator==` must be a refinement of `Pred` as explained [here](https://eel.is/c++draft/unord.req#general-237)?

pdimov:
That's a misunderstanding; I meant this part: `Notes:;;`, not the text. Note the semicolons instead of colons.

Re transfer, feel free to transfer the repo to me, and I will transfer it into boostorg.

Issue #154, opened 2022-10-31
Missing constructors, LWG Issue 2713


cmazakas:
See: https://cplusplus.github.io/LWG/issue2713

This is a blocker for the deduction guides mentioned in https://github.com/boostorg/unordered/issues/149

PR #160, opened 2022-11-03
Implement missing constructors outlined by LWG issue 2713

cmazakas:
Closes https://github.com/boostorg/unordered/issues/154

### DIFF
--- a/include/boost/unordered/unordered_map.hpp
+++ b/include/boost/unordered/unordered_map.hpp
@@ -123,6 +123,9 @@ namespace boost {
 
       explicit unordered_map(size_type, const hasher&, const allocator_type&);
 
+      template <class InputIterator>
+      unordered_map(InputIterator, InputIterator, const allocator_type&);
+
       template <class InputIt>
       unordered_map(InputIt, InputIt, size_type, const allocator_type&);
 
@@ -131,6 +134,8 @@ namespace boost {
         InputIt, InputIt, size_type, const hasher&, const allocator_type&);
 
 #if !defined(BOOST_NO_CXX11_HDR_INITIALIZER_LIST)
+      unordered_map(std::initializer_list<value_type>, const allocator_type&);
+
       unordered_map(
         std::initializer_list<value_type>, size_type, const allocator_type&);
 
@@ -1076,6 +1081,9 @@ namespace boost {
       explicit unordered_multimap(
         size_type, const hasher&, const allocator_type&);
 
+      template <class InputIterator>
+      unordered_multimap(InputIterator, InputIterator, const allocator_type&);
+
       template <class InputIt>
       unordered_multimap(InputIt, InputIt, size_type, const allocator_type&);
 
@@ -1084,6 +1092,9 @@ namespace boost {
         InputIt, InputIt, size_type, const hasher&, const allocator_type&);
 
 #if !defined(BOOST_NO_CXX11_HDR_INITIALIZER_LIST)
+      unordered_multimap(
+        std::initializer_list<value_type>, const allocator_type&);
+
       unordered_multimap(
         std::initializer_list<value_type>, size_type, const allocator_type&);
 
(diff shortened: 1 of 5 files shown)


include/boost/unordered/unordered_map.hpp
```diff
@@ -1751,6 +1762,16 @@ namespace boost {
     {
     }
 
+    template <class K, class T, class H, class P, class A>
+    template <class InputIterator>
+    unordered_map<K, T, H, P, A>::unordered_map(
+      InputIterator f, InputIterator l, const allocator_type& a)
+        : table_(boost::unordered::detail::initial_size(f, l, 0), hasher(),
```
pdimov, line 1769: This passes 0 for the initial bucket count...

include/boost/unordered/unordered_map.hpp
```diff
@@ -1773,6 +1794,16 @@ namespace boost {
 
 #if !defined(BOOST_NO_CXX11_HDR_INITIALIZER_LIST)
 
+    template <class K, class T, class H, class P, class A>
+    unordered_map<K, T, H, P, A>::unordered_map(
+      std::initializer_list<value_type> list, const allocator_type& a)
+        : table_(boost::unordered::detail::initial_size(
+                   list.begin(), list.end(), detail::default_bucket_count),
```
pdimov, line 1801: ... but this passes detail::default_bucket_count. Is this intended?

pdimov: The other two places both use detail::default_bucket_count, so probably not.

test/unordered/constructor_tests.cpp
```diff
@@ -317,66 +329,208 @@ namespace constructor_tests {
...
-    std::initializer_list<typename T::value_type> list;
+    typedef typename T::value_type value_type;
+
+    std::initializer_list<value_type> list;
+
+    test::random_values<T> v(3, generator);
+    std::vector<value_type> vec(v.begin(), v.end());
+    BOOST_ASSERT(vec.size() >= 3);
+
+    std::initializer_list<value_type> nonempty_list{vec[0], vec[1], vec[2]};
```
pdimov, line 340: This looks fishy to me. The backing array of the initializer_list is destroyed at the end of the full expression.

test/unordered/constructor_tests.cpp
```diff
@@ -317,66 +329,208 @@ namespace constructor_tests {
...
+        T x(list);
+        BOOST_TEST(x.empty());
+        BOOST_TEST_EQ(x.bucket_count(), 0u);
+        BOOST_TEST(test::equivalent(x.hash_function(), hf));
+        BOOST_TEST(test::equivalent(x.key_eq(), eq));
+        BOOST_TEST(test::equivalent(x.get_allocator(), al));
+      }
+
+      {
+        T x(nonempty_list);
```
pdimov, line 361: This should probably be `T x{ vec[0], vec[1], vec[2] };` instead.

test/unordered/constructor_tests.cpp
cmazakas: Hmm, that's a good question.

It seems to pass both sanitizers and checks for the values. For example, we still pass the check where we compare the container's contents to `expected`.

test/unordered/constructor_tests.cpp
cmazakas: This is a nice and simple refactor; will do!

include/boost/unordered/unordered_map.hpp
cmazakas: Good catch, we should be using `default_bucket_count` everywhere.

test/unordered/constructor_tests.cpp
pdimov: Yeah, declaring a variable of type initializer_list does extend the lifetime of the temporary backing array, https://eel.is/c++draft/dcl.init.list#6. So it's fine.

pdimov (approved):

pdimov:
This is going into 1.81, right?

joaquintides:
> This is going into 1.81, right?

Let's decide, because if it goes in 1.81 I have to update the docs. I'm for it.

pdimov:
We still have a few days, but you need to review and merge this to develop before I can merge it to master.

joaquintides (approved):
Hi Chris, I think the `boost::unordered::detail::default_bucket_count ` construct should be mimicked in foa with a dedicated, separate `boost::unordered::detail::foa::default_bucket_count`. As this is not directly related to the PR at hand, I'm approving but please keep this in your backlog. Thank you!

PR #161, opened 2022-11-07
Complete deduction tests, implement guides for all containers

cmazakas:
Closes https://github.com/boostorg/unordered/issues/149

### DIFF
--- a/include/boost/unordered/unordered_flat_map.hpp
+++ b/include/boost/unordered/unordered_flat_map.hpp
@@ -76,9 +76,9 @@ namespace boost {
       using init_type = typename map_types::init_type;
       using size_type = std::size_t;
       using difference_type = std::ptrdiff_t;
-      using hasher = Hash;
-      using key_equal = KeyEqual;
-      using allocator_type = Allocator;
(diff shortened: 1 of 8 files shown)


include/boost/unordered/unordered_flat_map.hpp
```diff
@@ -628,6 +628,104 @@ namespace boost {
...
+    namespace detail {
+      template <typename T>
+      using iter_key_t =
+        typename std::iterator_traits<T>::value_type::first_type;
+      template <typename T>
+      using iter_val_t =
+        typename std::iterator_traits<T>::value_type::second_type;
+      template <typename T>
+      using iter_to_alloc_t =
+        typename std::pair<iter_key_t<T> const, iter_val_t<T> >;
```
pdimov, line 642: I don't think `typename` is needed here.

joaquintides:
Why's the `boost::type_identity` trick only used for maps and not for sets?

pdimov:
This passed CI; should I merge it to develop and master? We can fix the remaining things later.

joaquintides:
I'm OK with that. Docs should be updated --I can do that, but, again, not this week

Issue #151, opened 2022-10-26
prime_fmod.hpp:120 triggers a runtime check failure #1 on msvc


mpalet:
detail\prime_fmod.hpp:120 triggers a runtime check failure #1 on msvc (Run-Time Check Failure #1 - A cast to a smaller data type has caused a loss of data.) 

Looks intentional due to the cast of boost::uint32_t(hash). 

It could be guarded somehow to not trigger this runtime failure when using msvc compiler.
Microsoft suggests the following:

Run-Time Check Failure #1 - A cast to a smaller data type has caused a loss of data.  If this was intentional, you should mask the source of the cast with the appropriate bitmask.  For example:  
	char c = (i & 0xFF);

cmazakas:
Hey there, I was able to replicate your issue locally. Unfortunately, it seems like you're using `/RTCc` which I found described here: https://learn.microsoft.com/en-us/cpp/build/reference/rtc-run-time-error-checks?view=msvc-170#arguments

The thing to note here is their statement:
> Because /RTCc rejects code that conforms to the standard, it's not supported by the C++ Standard Library.

So, this seems like a non-standard extension to run on code. In general, we only aim to support standards-conforming code.

To accommodate this use-case regardless, I did look into disabling the checks internally using:
```cpp
#pragma runtime_checks( "", off )
/* runtime checks are off in this region */
#pragma runtime_checks( "", restore )
```

around the problematic lines.

Unfortunately, it seems like msvc's pragmas don't like working with member functions of class templates. The compiler accepted where the pragmas were but did nothing to suppress the checks.

If you're able to tell me where to place the pragmas to disable the checks, I can add them but until then, I'm afraid there's nothing I can do.

greenkalx:
I've run into this too with our projects, migrating from boost_1_65_1 to boost_1_80_0.
(Visual Studio 2017 C++11 to VS2022 C++17)

For now, I am disabling in the project settings:
`properties > c/c++ > code generation > smaller type check /RTC1`
In the Visual Studio .vcxproj it is labelled:
`<SmallerTypeCheck>true</SmallerTypeCheck>`

We have it enabled in 300 projects for Debug builds only, not Release.
It's not triggered in our 40+ unit test projects.

`#pragma runtime_checks` is not working here either.
Should we file a MSVC compiler bug?

Does static_cast avoid it? I'll try tomorrow..

greenkalx:
Thank you @cmazakas, look forward to having this with 1.81.

[...]

PR #165, opened 2022-11-19
Add msvc RTC to select test targets

cmazakas:
Fixes https://github.com/boostorg/unordered/issues/151

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,11 +217,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { toolset: msvc-14.0, cxxstd: '14,latest',       addrmd: '32,64', os: windows-2019 }
-          - { toolset: msvc-14.2, cxxstd: '14,17,20,latest', addrmd: '32,64', os: windows-2019 }
-          - { toolset: msvc-14.3, cxxstd: '14,17,20,latest', addrmd: '32,64', os: windows-2022 }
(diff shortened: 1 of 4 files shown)


pdimov:
As discussed on Slack, instead of changing the Jamfile, /RTCc and /arch:IA32 need to be separate CI jobs.

.github/workflows/ci.yml
```diff
@@ -250,6 +252,9 @@ jobs:
           B2_TOOLSET: ${{matrix.toolset}}
           B2_CXXSTD: ${{matrix.cxxstd}}
           B2_ADDRESS_MODEL: ${{matrix.addrmd}}
+          B2_CXXFLAGS: ${{ (matrix.toolset == 'msvc-14.3' && matrix.cxxstd == '14' && ( matrix.addrmd == '64' && '/RTCc' || '"/RTCc /arch:IA32"')) || '' }}
```
pdimov, line 255: You need to add `cxxflags` and `defines` to the matrix, instead of hardcoding checks here.

test/unordered/prime_fmod_tests.cpp
```diff
@@ -155,7 +154,7 @@ void get_remainder_test()
 
   for (std::size_t i = 0; i < 1000000u; ++i) {
     boost::uint64_t f = rng();
-    boost::uint32_t d = static_cast<uint32_t>(rng());
+    boost::uint32_t d = rng() & 0xffffffffu;
```
pdimov, line 157: This probably still needs the static_cast.

pdimov (approved):

PR #172, opened 2022-12-01
feature/foa fast copy

joaquintides:
When copy constructing or copy assigning an unordered flat container, if the source and target bucket arrays have the same size, then:
* The metadata array is copied via `memcpy`.
* If `value_type` is trivially copy assignable, a single `memcpy` call is used to copy all the elements;
* else, elements are copied one by one but directly into place, i.e. bypassing the standard insertion algorithm.

This should speed up copy/copy assignment by an order of magnitude under the right conditions (haven't measured), and will make us look good in the related test on Martin's benchmarks.

### DIFF
--- a/include/boost/unordered/detail/foa.hpp
+++ b/include/boost/unordered/detail/foa.hpp
@@ -1533,8 +1533,13 @@ private:
 
   void copy_elements_array_from(const table& x,std::true_type /* -> memcpy */)
   {
+    /* reinterpret_cast: GCC may complain about value_type not being trivially
+     * copy-assignable when we're relying on trivial copy constructibility.
+     */
     std::memcpy(
-      arrays.elements,x.arrays.elements,x.capacity()*sizeof(value_type));
+      reinterpret_cast<unsigned char*>(arrays.elements),
+      reinterpret_cast<unsigned char*>(x.arrays.elements),
+      x.capacity()*sizeof(value_type));
   }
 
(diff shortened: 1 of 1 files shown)


pdimov (approved):
Looks good to me at a quick glance, but I haven't checked whether our copy tests are extensive enough to exercise all these cases.

joaquintides:
Codecov says all the new code is exercised:
https://app.codecov.io/gh/boostorg/unordered/blob/feature%2Ffoa_fast_copy/include/boost/unordered/detail/foa.hpp

cmazakas (approved):
I'm not sure how else to ensure a failing test case aside from adding a static assertion somewhere in the copy tests that we're using a trivially copyable type but this otherwise seems great to me.

joaquintides:
Shall we merge now or do we keep develop untouched till Boost 1.81 just in case some last.minute fix is needed=

pdimov:
Maybe just wait, it's two weeks.

pavel-odintsov:
That's impressive stuff! I think you remember my weird questions about ways to copy structure as fast as possible :)

joaquintides:
Hi Pavel, thank you! Are you going to give `boost::unordered_flat_map` a try with FastNetMon?

pavel-odintsov:
According to Joaquín M López Muñoz this thing will be part of Boost 1.82.

PR #173, opened 2022-12-03
Exception Guarantees

cmazakas:
Update assignment operators and swap implementation to require noexcept swapability of the Hash, KeyEqual params. Correspondingly update tests as needed.


include/boost/unordered/detail/foa.hpp
```diff
@@ -1217,13 +1217,22 @@ table:empty_value<Hash,0>,empty_value<Pred,1>,empty_value<Allocator,2>
...
     static constexpr auto pocca=
       alloc_traits::propagate_on_container_copy_assignment::value;
 
     if(this!=std::addressof(x)){
+      hasher    tmp_h=x.h();
+      key_equal tmp_p=x.pred();
+
+      using std::swap;
+      swap(h(),tmp_h);
+      swap(pred(),tmp_p);
```
pdimov, line 1232: Explanatory comments are missing here. "// can throw" on the tmp lines, then "// swap operations don't throw, so we know we have consistent hasher and key_equal at this point." Or even more verbose.

test/exception/swap_exception_tests.cpp
```diff
@@ -4,15 +4,40 @@
...
+namespace boost {
+  void assertion_failed(
+    char const* expr, char const* function, char const* file, long line)
+  {
+    (void)expr;
+    (void)function;
+    (void)file;
+    (void)line;
+
+    throw std::runtime_error("rawr!");
```
pdimov, line 26: Not very professional. :-)

test/exception/swap_exception_tests.cpp
```diff
@@ -140,11 +160,132 @@ template <class T> struct swap_test4 : swap_base<T>
...
+  std::equal_to<int>, test::allocator1<std::pair<int const, int> > >
+  unordered_multimap;
+
+#define SWAP_CONTAINER_SEQ                                                     \
+  (unordered_set)(unordered_map)(unordered_multiset)(unordered_multimap)
+#endif
+
+// FOA containers deliberately choose to not offer the strong exception
+// guarantee so we can't reliably test what happens if swapping one of the data
+// members throws as the basic guarantee doesn't uphold invariants
```
pdimov, line 274: The basic guarantee upholds invariants.

joaquintides (commented):
* Why are you using `BOOST_STATIC_ASSERT` instead of plain `static_assert`?
* Move assignment: you're swapping `h`  and `pred`, which feels incorrect to me (the source container need not keep target objects). I'd rather do:
```
hasher    tmp_h=std::move(x.h());
key_equal tmp_p=std::move(x.pred());

using std::swap;
swap(h(),tmp_h);
swap(pred(),tmp_p);
```
* You've replaced `swap_if<pocs>(al(),x.al())` with `swap(al(),x.al())`: I think this is incorrect, as the standard [says](https://eel.is/c++draft/container.requirements#container.reqmts-65) that allocators must we swappable only if `allocator_traits<allocator_type>::propagate_on_container_swap::value` is `true`.
* Docs should be updated to, at least, reflect the new `noexcept` conditions.

include/boost/unordered/detail/foa.hpp
cmazakas: Hopefully these comments offer a bit more clarity.

cmazakas:
> Why are you using BOOST_STATIC_ASSERT instead of plain static_assert?

It was so that I didn't have to come up with an error message. I think only C++17 allows plain `static_assert(condition);`.

> Move assignment: you're swapping h and pred, which feels incorrect to me (the source container need not keep target objects). I'd rather do:

This becomes at-odds with our new `noexcept()` requirements, as I've found via the failing `noexcept_tests`. Technically, this is an odd-but-valid way of implementing noexcept move passively from our noexcept swap without imposing any new requirements on the Hash, KeyEqual params.

> Docs should be updated to, at least, reflect the new noexcept conditions.
Done!

joaquintides:
* `BOOST_STATIC_ASSERT`: I still prefer plain `static_assert` since C++11 is assumed, but this is a stylistic issue anyway. You're missing the `#include <boost/static_assert.hpp>`  in `foa.hpp`. 
* Move assignment: I'd rather keep the old `noexcept` specification because this is what the standard explicitly has, and use the `swap` trick only as an internal way to guarantee basic safety (so, doing first a move-assignment of the objects to local variables). We can change the `noexcept` declaration when the standard does, which presumable will be in the direction of requiring non-throwing move assignability of `hash` and `pred` (making our trick superfluous).
* Docs: the `noexcept`s in the _synopses_ are not in sync with those in the detailed explanations for `swap` and move assignment. If the _current_ implementation of move assignment in this PR prevails, then the statement _"move-assigns the hash function and predicate from other"_ is incorrect.

include/boost/unordered/detail/foa.hpp
```diff
@@ -67,6 +67,12 @@
   }while(0)
 #endif
 
+#define BOOST_UNORDERED_STATIC_ASSERT_HASH_PRED(Hash, Pred)                    \
+  static_assert(boost::is_nothrow_swappable<Hash>::value,                      \
+    "Template parameter Hash is required to be nothrow Swappable.");           \
+  static_assert(boost::is_nothrow_swappable<Pred>::value,                      \
```
joaquintides, line 73: s/Swappable/swappable

joaquintides (approved):

pdimov (approved):

PR #174, opened 2022-12-05
feature/pr165 refactor

joaquintides:
Refactoring of [PR165](https://github.com/boostorg/unordered/pull/165) introducing a [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md)-inspired `narrow_cast` that internally takes care of making VS run-time checks happy.

### DIFF
--- a/include/boost/unordered/detail/foa.hpp
+++ b/include/boost/unordered/detail/foa.hpp
@@ -23,6 +23,7 @@
 #include <boost/predef.h>
 #include <boost/type_traits/has_trivial_copy.hpp>
 #include <boost/type_traits/is_nothrow_swappable.hpp>
+#include <boost/unordered/detail/narrow_cast.hpp>
 #include <boost/unordered/detail/xmx.hpp>
 #include <boost/unordered/hash_traits.hpp>
 #include <climits>
@@ -269,20 +270,12 @@ private:
       0xF8F8F8F8u,0xF9F9F9F9u,0xFAFAFAFAu,0xFBFBFBFBu,0xFCFCFCFCu,0xFDFDFDFDu,0xFEFEFEFEu,0xFFFFFFFFu,
     };
 
-#if defined(__MSVC_RUNTIME_CHECKS)
-    return (int)word[hash&0xffu];
-#else
-    return (int)word[(unsigned char)hash];
-#endif
+    return (int)word[narrow_cast<unsigned char>(hash)];
   }
 
   inline static unsigned char reduced_hash(std::size_t hash)
   {
-#if defined(__MSVC_RUNTIME_CHECKS)
-    return match_word(hash)&0xffu;
-#else
-    return (unsigned char)match_word(hash);
-#endif
+    return narrow_cast<unsigned char>(match_word(hash));
   }
 
   inline unsigned char& at(std::size_t pos)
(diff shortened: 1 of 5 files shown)


pdimov:
What's the contract/specification of narrow_cast?

All that ifdefing on rvalue references doesn't seem necessary to me.

joaquintides:
I just copied [this implementation](https://github.com/microsoft/GSL/blob/main/include/gsl/util#L93-L101) from GSL. Chris is also in favor of simplifying the signature, and has run Godbolt experiments showing that codegen is the same, so [here](https://github.com/boostorg/unordered/pull/174/commits/100598e285b3e7269bdd63079ee351f15211560c) we go.

pdimov:
Yeah, OK, but what's the specification of narrow_cast? What are the requirements on From and To?

joaquintides:
They must be integral types with `sizeof(From)>=sizeof(To)`.

pdimov:
Can either of them be signed? Can they differ in signedness? (sizeof isn't quite the right test, it's more correct to test numeric_limits<T>::digits, but I get what you mean.)

joaquintides:
They can differein signedness, though we're only using it with same-signed integral types.

pdimov:
In that case, we need the correct static asserts in narrow_cast, and a test for it exercising all the possibilities. (Visual inspection isn't enough for me to be sure that it works correctly for all the various combinations of signed/unsigned types.)

joaquintides:
I've just expanded the static asserts as indicated. As for testing, the following:
```cpp
#include <boost/config.hpp>
#include <boost/static_assert.hpp>
#include <boost/type_traits/is_integral.hpp>

template<typename To,typename From>
BOOST_CONSTEXPR To narrow_cast(From x) BOOST_NOEXCEPT
{
  BOOST_STATIC_ASSERT_MSG(
    boost::is_integral<From>::value&&
    boost::is_integral<To>::value&&
    sizeof(From)>=sizeof(To),
    "From and To must be integral types with sizeof(From)>=sizeof(To)");

  return static_cast<To>(
    x
    &static_cast<To>(~static_cast<To>(0))
  );
}

#include <cassert>
#include <cstdint>
#include <limits>

template<typename From,typename To>
void test_narrow_cast()
{
  for(From n=std::numeric_limits<From>::min();
      n<std::numeric_limits<From>::max();
      ++n){
    assert(narrow_cast<To>(n)==static_cast<To>(n));
  }
}

int main()
{
  test_narrow_cast<signed int,unsigned char>();
  test_narrow_cast<unsigned int,signed char>();
  test_narrow_cast<std::uint64_t,std::uint32_t>();
  test_narrow_cast<std::uint64_t,signed short>();
}
```
[works fine](http://coliru.stacked-crooked.com/a/91934afe689b8b12). Note that the testing code exercices the `__MSVC_RUNTIME_CHECKS` branch, which is the more contentious (otherwise, `narrow_cast` is just a `static_cast`). I don't think this testing needs to be included as part of Boost.Unordered, as the component is an internal one, and correctness surely can be formally proved by someone fluent in standardese.

cmazakas:
I wouldn't mind adding the test file. It'd be simple enough and we'd have some good coverage over it. We don't even need to do exhaustive checks, just as many different edge-cases as we can think of.

pdimov:
The problem with doing the test like this (checking against `static_cast`) is that it will do nothing without `/RTCc`, and abort with `/RTCc`. I would just check a few reference values, e.g. `BOOST_TEST_EQ( narrow_cast<signed char>( -1 ), (signed char)-1 );` and similar.

It's better if the static asserts are separated; this way there's information as to which condition isn't met. While some recent compilers actually decompose the condition and tell you which part failed, far from all do. (And there would be no need for _MSG in that case, because it would just repeat the condition, except in English.)

cmazakas:
I think the tests are good. I like how narrow_cast() came out and how it looks in the code so I'm okay with all the changes. Let me know what you guys think!

cmazakas (approved):

pdimov (approved):

joaquintides:
Looks all good to me. Thans for testing this thoroughly,

PR #176, opened 2022-12-12
feature/foa fast copy fix

joaquintides:
Restores `boost::unordered_flat_[map|set]`'s allocator awareness by restricting `std::memcpy`-based element copy to `std::allocator` and allocators without a suitable  `construct` memfun.

### DIFF
--- a/include/boost/unordered/detail/foa.hpp
+++ b/include/boost/unordered/detail/foa.hpp
@@ -31,6 +31,7 @@
 #include <cstring>
 #include <iterator>
 #include <limits>
+#include <memory>
 #include <tuple>
 #include <type_traits>
 #include <utility>
@@ -1025,6 +1026,47 @@ inline void prefetch(const void* p)
 
 struct try_emplace_args_t{};
 
+template<typename Allocator>
+struct is_std_allocator:std::false_type{};
+
+template<typename T>
+struct is_std_allocator<std::allocator<T>>:std::true_type{};
+
+/* std::allocator::construct marked as deprecated */
+#if defined(_LIBCPP_SUPPRESS_DEPRECATED_PUSH)
+_LIBCPP_SUPPRESS_DEPRECATED_PUSH
+#elif defined(_STL_DISABLE_DEPRECATED_WARNING)
+_STL_DISABLE_DEPRECATED_WARNING
+#elif defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable:4996)
+#endif
+
+template<typename Allocator,typename Ptr,typename... Args>
+struct alloc_has_construct
+{
+private:
+  template<typename Allocator2>
+  static decltype(
+    std::declval<Allocator2&>().construct(
+      std::declval<Ptr>(),std::declval<Args&&>()...),
+    std::true_type{}
+  ) check(int);
+
+  template<typename> static std::false_type check(...);
+
+public:
+  static constexpr bool value=decltype(check<Allocator>(0))::value;
+};
+
+#if defined(_LIBCPP_SUPPRESS_DEPRECATED_POP)
+_LIBCPP_SUPPRESS_DEPRECATED_POP
+#elif defined(_STL_RESTORE_DEPRECATED_WARNING)
+_STL_RESTORE_DEPRECATED_WARNING
+#elif defined(_MSC_VER)
+#pragma warning(pop)
+#endif
+
 #if defined(BOOST_GCC)
 /* GCC's -Wshadow triggers at scenarios like this: 
  *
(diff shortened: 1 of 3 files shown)


pdimov:
Why are the warning restore pragmas not in reverse order of the suppression ones?

test/objects/test.hpp
```diff
@@ -236,10 +236,10 @@ namespace test {
         result = x;
         break;
       case 2:
-        result = x * 7;
+        result = static_cast<int>(static_cast<boost::long_long_type>(x) * 7);
```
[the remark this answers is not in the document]
cmazakas: That is a good question. Technically, this one didn't trigger ubsan's overflow checks.

Currently, only `x * 256` did but I figured that if I was doing one, I'd also do the other just to be safe.

The test suite seems to pass fine in the case of `x * 7` so if you'd like me to revert this change, it's no problem.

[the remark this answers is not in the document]
pdimov: I remember that when I initially fixed the ubsan failures, these two did not trigger. Does a newly added test cause an overflow here?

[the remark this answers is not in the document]
pdimov: Either way, this should probably be changed in the same way as the two previous functions, for consistency.

include/boost/unordered/detail/foa.hpp
```diff
@@ -1023,6 +1024,47 @@ inline void prefetch(const void* p)
...
+_LIBCPP_SUPPRESS_DEPRECATED_PUSH
+#elif defined(_STL_DISABLE_DEPRECATED_WARNING)
+_STL_DISABLE_DEPRECATED_WARNING
+#elif defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable:4996)
+#endif
+
+template<typename Allocator,typename Ptr,typename... Args>
+struct alloc_has_construct
```
glenfe, line 1044: Maybe the better thing to do is we instead enhance `boost::allocator_construct_n()` to call `memcpy()` when `alloc_has_construct` is `false` and `T` is trivially copyable.

cmazakas: @joaquintides Personally, I'd be interested in seeing this. How do you feel? I don't mind holding back the PR to wait for an implementation.

glenfe: No need to hold back the PR. :) Can always refactor later.

pdimov: I'm not sure this will help here, because the bucket array can have "holes" in it, unconstructed elements, but is still copied with a single memcpy. (Not 100% positive though.)

pdimov: Maybe just expose `allocator_has_construct` as a supported, user-facing, primitive.

joaquintides:
> Why are the warning restore pragmas not in reverse order of the suppression ones?

I opted for having those pragmas be mutually exclusive (note the `#elif`s), hence restore order can't be reverse (some branches are true simultaneously).

include/boost/unordered/detail/foa.hpp
joaquintides: It's as Peter says, the `std::memcpy` call goes through an array with uninitialized portions. Having `allocator_has_construct` exposed would save us some some code duplication.

include/boost/unordered/detail/foa.hpp
```diff
@@ -1559,6 +1601,9 @@ table:empty_value<Hash,0>,empty_value<Pred,1>,empty_value<Allocator,2>
 #else
         std::is_trivially_copy_constructible<value_type>::value
 #endif
+        &&(
+          is_std_allocator<Allocator>::value||
```
glenfe, line 1605: What about users who specialize `std::allocator`?

joaquintides: Well, I guess they shouldn't :-) FWIW, stdlib implementors also assume that `std::allocator` does not have a fancy `construct` regardless of specializations:

https://github.com/llvm-mirror/libcxx/blob/master/include/memory#L1670-L1701

The way I see it, `std::memcpy` is too good not to use for the sake of corner cases such as you mention, which are in any case better served by proper user-defined allocators.

joaquintides: libstdc++ also assumes `std::allocator` is not to be user specialized:
https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_uninitialized.h#L334-L375

pdimov (approved):

Issue #150, opened 2022-10-26
No implementation of P2363


cmazakas:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2363r3.html

iwubcode:
Yes please.  I would love to see this in time for 1.81!

cmazakas:
This has been implemented and is on develop. I'm going to close the issue.

iwubcode:
Wonderful news.  Thank you!

PR #177, opened 2022-12-14
Feature/p2363

cmazakas:
Implement the rest of the transparent operations as specified by: http://wg21.link/P2363

Addresses https://github.com/boostorg/unordered/issues/150

### DIFF
--- a/doc/unordered/unordered_flat_map.adoc
+++ b/doc/unordered/unordered_flat_map.adoc
@@ -134,23 +134,31 @@ namespace boost {
       std::pair<iterator, bool> xref:#unordered_flat_map_try_emplace[try_emplace](const key_type& k, Args&&... args);
     template<class... Args>
       std::pair<iterator, bool> xref:#unordered_flat_map_try_emplace[try_emplace](key_type&& k, Args&&... args);
+    template<class K, class... Args>
+      std::pair<iterator, bool> xref:#unordered_flat_map_try_emplace[try_emplace](K&& k, Args&&... args);
     template<class... Args>
       iterator xref:#unordered_flat_map_try_emplace_with_hint[try_emplace](const_iterator hint, const key_type& k, Args&&... args);
     template<class... Args>
       iterator xref:#unordered_flat_map_try_emplace_with_hint[try_emplace](const_iterator hint, key_type&& k, Args&&... args);
+    template<class K, class... Args>
+      iterator xref:#unordered_flat_map_try_emplace_with_hint[try_emplace](const_iterator hint, K&& k, Args&&... args);
     template<class M>
       std::pair<iterator, bool> xref:#unordered_flat_map_insert_or_assign[insert_or_assign](const key_type& k, M&& obj);
     template<class M>
       std::pair<iterator, bool> xref:#unordered_flat_map_insert_or_assign[insert_or_assign](key_type&& k, M&& obj);
+    template<class K, class M>
+      std::pair<iterator, bool> xref:#unordered_flat_map_insert_or_assign[insert_or_assign](K&& k, M&& obj);
     template<class M>
       iterator xref:#unordered_flat_map_insert_or_assign_with_hint[insert_or_assign](const_iterator hint, const key_type& k, M&& obj);
     template<class M>
       iterator xref:#unordered_flat_map_insert_or_assign_with_hint[insert_or_assign](const_iterator hint, key_type&& k, M&& obj);
+    template<class K, class M>
+      iterator xref:#unordered_flat_map_insert_or_assign_with_hint[insert_or_assign](const_iterator hint, K&& k, M&& obj);
 
     void      xref:#unordered_flat_map_erase_by_position[erase](iterator position);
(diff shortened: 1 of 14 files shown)


joaquintides (changes requested):
* Additional to the in-code comments, the PR lacks documentation. I can do this over the weekend.

include/boost/unordered/detail/type_traits.hpp
```diff
@@ -53,10 +53,10 @@ namespace boost {
       {
       };
 
-      template <class, class A, class B> struct are_transparent
+      template <class, class Hash, class KeyEqual> struct are_transparent
       {
         static bool const value =
-          is_transparent<A>::value && is_transparent<B>::value;
+          is_transparent<Hash>::value && is_transparent<KeyEqual>::value;
```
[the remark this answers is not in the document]
cmazakas: The SFINAE for the transparent member functions only works when they're dependent on the template parameter so we use a dummy here to keep the check dependent.

include/boost/unordered/detail/implementation.hpp
```diff
@@ -1477,6 +1477,33 @@ namespace boost {
...
+          return a.release();
+        }
+
+        template <typename T, typename V, typename Alloc, typename Key>
+        inline typename boost::allocator_pointer<Alloc>::type
+        try_construct_node(
+          std::pair<T const, V>*, Alloc& alloc, BOOST_FWD_REF(Key) k)
+        {
+          return construct_node_pair(alloc, boost::forward<Key>(k));
+        }
```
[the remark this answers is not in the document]
cmazakas: Yeah, the `try_` prefix came from the fact that these functions are _only_ being used in `try_emplace_unique`.

Good call on the first overload not delegating properly as it should.

`construct_node_or_node_pair` is fine. Maybe something like `construct_node_from_key_only` could also work.

joaquintides:
While updating the docs, I noticed that `unordered_map::try_emplace` (unlike `emplace`) does not add overloads up to 10 arguments to emulate variadic args, despite what the reference says. Note related to PR, you may want to file an issue.

joaquintides (changes requested):
There's a bug with the P2363 spec that shows in our implementation of it:
```cpp
struct transparent_hash
{
  using is_transparent=void;

  template<typename T>
  std::size_t operator()(const T& x)const
  {
    return boost::hash<T>()(x);
  }
};

using transparent_set=boost::unordered_flat_set<int,transparent_hash,std::equal_to<>>;
...
transparent_set ts1,ts2;
ts1.insert(ts2.begin(),ts2.end());
```
The call to `insert` is ambiguous (could be heterogeneous-lookup `insert` with a hint or range `insert`).

joaquintides (approved):

doc/unordered/unordered_flat_map.adoc
```diff
@@ -938,17 +971,25 @@ If there is an element with key `k`, then it is updated by assigning `boost::for
...
+           std::forward_as_tuple(boost::forward<M>(obj)))
 ```
 
 [horizontal]
 Returns:;; The `bool` component of the return type is `true` if an insert took place. +
 +
 If an insert took place, then the iterator points to the newly inserted element. Otherwise, it points to the element with equivalent key.
 Throws:;; If an exception is thrown by an operation other than a call to `hasher` the function has no effect.
-Notes:;; Can invalidate iterators pointers and references, but only if the insert causes the load to be greater than the maximum load. +
+Notes:;; Can invalidate iterators pointers and references, but only if the insert causes the load to be greater than the maximum load.  +
```
pdimov, line 990: Is `:;;` correct here? Shouldn't it be `:::`?

doc/unordered/unordered_multimap.adoc
```diff
@@ -1330,11 +1305,13 @@ Returns:;; The number of elements in bucket `n`.
 ==== bucket
 ```c++
 size_type bucket(const key_type& k) const;
+templat<class K> size_type bucket(const K& k) const;
```
pdimov, line 1308: typo: `templat`

doc/unordered/unordered_flat_map.adoc
joaquintides: This is what I found when I begin updating the docs (no ocurrence of `:::` overall), and it displays OK, but other than this I don't know. Maybe @cmazakas knows more about Asciidoc syntax to shed some light here.

doc/unordered/unordered_multimap.adoc
joaquintides: Fixed

doc/unordered/unordered_flat_map.adoc
pdimov: Maybe `;;` is an alternative token?

pdimov: Yes, it is. https://docs.asciidoctor.org/asciidoc/latest/lists/description/

pdimov (approved):

PR #182, opened 2023-02-13
Feature/unordered node map docs

joaquintides:
This is complete except the benchmarks, which I'll extend to cover `unordered_node_map` in the following days.


doc/unordered/compliance.adoc
```diff
@@ -117,27 +117,29 @@ Variadic constructor arguments for `emplace` are only used when both
...
+* In general:
+  ** `begin()` is not constant-time.
+  ** `erase(iterator)` returns `void` instead of an iterator to the following element.
+  ** There is no API for bucket handling (except `bucket_count`).
+  ** The maximum load factor of the container is managed internally and can't be set by the user. The maximum load,
+     exposed through the public function `max_load`, may decrease on erasure under high-load conditions.
+* Flat containers (`boost::unordered_flat_set` and `boost::unordered_flat_map`):
+  ** `value_type` must be move-constructible.
+  ** Pointer stability is not kept under rehashing.
+  ** There is no API for node extraction/insertion
```
pdimov, line 145: A period seems to be missing here.

pdimov (approved):

doc/unordered/buckets.adoc
```diff
@@ -252,15 +254,16 @@ xref:#rationale_boostunordered_multiset_and_boostunordered_multimap[correspondin
...
+`boost:unordered_flat_set`/`unordered_node_set`.
 
 
 [#img-foa-layout]
 .Open-addressing layout used by Boost.Unordered.
 image::foa.png[align=center]
 
-As with all open-addressing containers, elements are stored directly in the bucket array.
+As with all open-addressing containers, elements (or element nodes in the case of
+`boost::unordered_node_map` and `boost::unordered_node_set`) are stored directly in the bucket array.
```
cmazakas, line 266: Not sure I like the term" element nodes".

What about something more like this:
> As with all open-addressing containers, elements (or pointers thereto) are stored directly in the bucket array.

doc/unordered/compliance.adoc
```diff
@@ -117,27 +117,29 @@ Variadic constructor arguments for `emplace` are only used when both
 rvalue references and variadic template parameters are available.
 Otherwise `emplace` can only take up to 10 constructors arguments.
 
-== Open-addressing containers: unordered_flat_set, unordered_flat_map
+== Open-addressing containers: unordered_flat_set/unordered_node_set, unordered_flat_map/unordered_node_map
```
cmazakas, line 120: These titles are getting a little unruly for me
Screenshot_20230213_121328

It spills over in my browser and requires scrolling to read the entire thing.

Instead, maybe we just trim this down to `Open-Addressing Containers` or something like that.

doc/unordered/buckets.adoc
joaquintides: Changed to "pointers to the element nodes": I think it's important that we retain the term "node" as we're constantly referring to "node-based containers".

doc/unordered/compliance.adoc
joaquintides: Done.

doc/unordered/benchmarks.adoc
```diff
@@ -278,13 +278,14 @@ max load factor 5
...
 |===
 
-== boost::unordered_flat_map
+== boost::unordered_(flat|node)_map
 
 All benchmarks were created using:
 
 * `https://abseil.io/docs/cpp/guides/container[absl::flat_hash_map^]<uint64_t, uint64_t>`
-* `boost::unordered_flat_map<uint64_t, uint64_t>`
 * `boost::unordered_map<uint64_t, uint64_t>`
```
cmazakas, line 286: Should we replace `boost::unordered_map` with `absl::node_hash_map` for our benchmarks here?

PR #184, opened 2023-02-24
Feature/SIMD config

joaquintides:
Though the changes are trivial, this feature opens up infinite possibilities for bikeshedding. Your opinion requested on the following:

* Names ok? `BOOST_UNORDERED_..._LITTLE_ENDIAN_NEON` could be `BOOST_UNORDERED_..._NEON`, but this allows us more granularity in case we eventually support big-endian Neon.
* Disabling, say Neon, does not guarantee that the non-SIMD variant will be used: SSE2 can still be automatically detected. Is there value in forcing non-SIMD?
* Do we _document_ this? I'd rather not do it for the moment being. FWIW, `BOOST_URL_USE_SSE2` is not documented either.


pdimov:
I don't see why one would want to specifically disable little-endian NEON. BOOST_UNORDERED_DISABLE_NEON should be enough.

joaquintides:
a) `BOOST_UNORDERED_DISABLE_NEON`/`BOOST_UNORDERED_ENABLE_LITTLE_ENDIAN_NEON` or b) `BOOST_UNORDERED_DISABLE_NEON`/`BOOST_UNORDERED_ENABLE_NEON`? Note that blindly enabling Neon on a big-endian Neon platform will lead to either UB or a compile-time error (not sure which, really).

pdimov:
BOOST_UNORDERED_DISABLE_NEON/BOOST_UNORDERED_ENABLE_NEON, I think. The user will be responsible for whatever happens, these are "manual override" buttons that should be needed very rarely.

include/boost/unordered/detail/foa.hpp
```diff
@@ -39,12 +39,24 @@
 #include <type_traits>
 #include <utility>
 
-#if defined(__SSE2__)||\
+#if !defined(BOOST_UNORDERED_DISABLE_SSE2)&& \
+    !defined(BOOST_UNORDERED_ENABLE_SSE2)
+#if defined(__SSE2__)|| \
     defined(_M_X64)||(defined(_M_IX86_FP)&&_M_IX86_FP>=2)
-#define BOOST_UNORDERED_SSE2
+#define BOOST_UNORDERED_ENABLE_SSE2
```
pdimov, line 46: I don't like mixing up macros defined by the user with macros defined by the library. ENABLE/DISABLE should only be set by the user and never by the library. We should retain our own macro and define it accordingly.

pdimov:
Perfect, thanks.

Are we sure we want to undefine our SSE2/Neon macros at the end? It might be useful for a user to know whether SSE2/Neon is used or not.

joaquintides:
I think it's best not to leak these macros because they can be Hyrum-abused: we may, for instance, complete forego SSE2 in the future in favor of a revolutionary SWAR technique, and we'd be force to keep computing `BOOST_UNORDERED_SSE2` to not break users who take advantage of it for their own SSE2-detection.

pdimov:
That's their problem. The intended use of this macro is to detect whether we use SSE2, not whether someone else should.

cmazakas:
> That's their problem. The intended use of this macro is to detect whether we use SSE2, not whether someone else should.

In general, I agree with this sentiment.

Users would probably be very interested in making sure _our_ code is using SIMD. If they do choose to use us for detection logic, that was always kind of dicey, especially if we agree to not publicly document the macros.

PR #185, opened 2023-02-27
Construct stack-locals using Allocator

cmazakas:


include/boost/unordered/unordered_flat_map.hpp
```diff
@@ -61,11 +66,26 @@ namespace boost {
...
         template <class A, class... Args>
-        static void construct(A& al, element_type* p, Args&&... args)
+        static void construct(A& al, init_type* p, Args&&... args)
         {
           boost::allocator_construct(al, p, std::forward<Args>(args)...);
         }
 
+        template <class A, class... Args>
+        static void construct(A& al, element_type* p, Args&&... args)
+        {
```
joaquintides, line 76: This looks inconsistent: why are you rebinding here and not in `construct(A& al, init_type* p, Args&&... args)`? I understand the contract is that the caller passes an appropriately rebound allocator.

joaquintides: Also, consider renaming `construct(A& al, element_type* p, Args&&... args)` and `destroy(A& al, element_type* p)` to `construct(A& al, value_type* p, Args&&... args)` and `destroy(A& al, value_type* p)`: for flat containers, `element_type` is a mere alias to `value_type`  so it looks nicer if `element_type` does not appear in the policy except for its definition (unlike in node containers, where `element_type` receives special treatment).

include/boost/unordered/unordered_flat_map.hpp
```diff
@@ -53,6 +53,11 @@ namespace boost {
           return kv.first;
         }
 
+        static moved_type move(init_type& x)
+        {
+          return {std::get<0>(std::move(x)), std::get<1>(std::move(x))};
+        }
```
joaquintides, line 59: Why not `{std::move(x.first), std::move(x.second)};`?

Same for other places where this construct appears.

include/boost/unordered/detail/foa.hpp
```diff
@@ -1129,6 +1129,25 @@ _STL_RESTORE_DEPRECATED_WARNING
...
+template <class T>
+union storage
+{
+  T t_;
+  storage(){}
+  ~storage(){}
+};
+
+template <class TypePolicy,class A,class T>
+struct drop_guard
```
joaquintides, line 1141: I know I asked for this to be defined outside `table`, but upon re-reviewing I think it makes more sense to rename this as `destroy_on_exit` and define it in-class alongside `clear_on_exit` and `erase_on_exit`.

include/boost/unordered/unordered_node_map.hpp
```diff
@@ -106,10 +123,11 @@ namespace boost {
...
           {
-            boost::allocator_deallocate(al,
-              boost::pointer_traits<
-                typename boost::allocator_pointer<A>::type>::pointer_to(*p->p),
-              1);
+            using pointer_type=typename boost::allocator_pointer<A>::type;
+            using pointer_traits=boost::pointer_traits<pointer_type>;
+
+            boost::allocator_deallocate(
+              al,pointer_traits::pointer_to(*(p->p)),1);
```
joaquintides, line 130: Why not just `boost::allocator_deallocate(al,p->p,1)`?

include/boost/unordered/unordered_node_map.hpp
```diff
@@ -118,16 +136,22 @@ namespace boost {
...
         template <class A> static void destroy(A& al, element_type* p) noexcept
         {
           if (p->p) {
-            destroy(al,p->p);
+            using pointer_type=typename boost::allocator_pointer<A>::type;
+            using pointer_traits=boost::pointer_traits<pointer_type>;
+
+            destroy(al, p->p);
+            boost::allocator_deallocate(
+              al,pointer_traits::pointer_to(*(p->p)), 1);
```
joaquintides, line 154: Why not just `boost::allocator_deallocate(al,p->p,1)`?

joaquintides: Also, is not `al` supposed to be bound to `element_type`?

joaquintides (commented):
Additionally to the inline comments:

* I think there's some confusion with `al` parameters in the `construct`/`destroy` overloads in the type policies; we should stick to one of these policies:
  * `al` is properly bound to the type passed (my preferred option).
  * `al` is bound to some fixed type (say `value_type`) everywhere and rebinding is done inside `construct`/`destroy` as appropriate.
* The explanation for type policies in https://github.com/boostorg/unordered/blob/fix/uses-allocator/include/boost/unordered/detail/foa.hpp#L1151-L1208 needs to be updated (type policies' functionality has been expanded).
* Consider attaching a link in the conversation to the benchmarks that show that performance is preserved.
* Consider adding a bullet to the release notes with this fix.

pdimov:
`construct` and `destroy` don't require rebinding if I'm not mistaken.

include/boost/unordered/unordered_flat_map.hpp
pdimov: That's not the same, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108952

pdimov: Ah wait, it's the same here. :-)

include/boost/unordered/detail/foa.hpp
```diff
@@ -1129,6 +1129,25 @@ _STL_RESTORE_DEPRECATED_WARNING
  */
 constexpr static float const mlf = 0.875f;
 
+template <class T>
+union storage
+{
+  T t_;
+  storage(){}
+  ~storage(){}
```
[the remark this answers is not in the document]
cmazakas: It seems to be required when the type has a non-trivial destructor. 

The following example demonstrates:
https://godbolt.org/z/1s5zP738b

include/boost/unordered/unordered_node_map.hpp
cmazakas: In our cases, the `element_type` stores a `value_type*` which we can't deallocate through if the Allocator is using fancy pointers.

I think it's easiest for us to store `value_type*` and convert from raw pointers to fancy pointers only in these certain instances.

cmazakas:
> `construct` and `destroy` don't require rebinding if I'm not mistaken.

Actually, good point; this is correct.

I looked into a bit more and yeah, an Allocator's value_type only matters in the case of allocate(), deallocate().

In these cases, our allocator is already bound to `value_type` which means, we don't need any rebinding at all!

include/boost/unordered/unordered_node_map.hpp
cmazakas: `al` shouldn't be bound to `element_type` because in all actuality, we're only allocating and deallocating the `type_policy::value_type`.

If we wanted to harden these measures, we should probably add:
```cpp
BOOST_STATIC_ASSERT(
  std::is_same<
    typename boost::allocator_value_type<A>::type,
    type_policy::value_type
  >::value);
```

include/boost/unordered/unordered_flat_map.hpp
cmazakas: Good catch. I was erroneously rebinding, likely as an artifact of early passes at this refactor.

The code has been updated such that it no longer needs to rebind at all. We know that our Allocator is properly bound to allocate, deallocate value_type. The construct(), destroy() methods are supposed to work with any and all types without the need for rebinding.

include/boost/unordered/detail/foa.hpp
```diff
@@ -1641,6 +1656,16 @@ table:empty_value<Hash,0>,empty_value<Pred,1>,empty_value<Allocator,2>
     bool           rollback_=false;
   };
 
+  template <class T>
+  struct destroy_on_exit
+  {
+    using type_policy=TypePolicy;
```
joaquintides, line 1662: Is this `using` needed?

include/boost/unordered/unordered_node_map.hpp
joaquintides: Yes, I agree. Thank you!

include/boost/unordered/unordered_node_map.hpp
joaquintides: I think it's ok as it's now, no need to go extra picky and do the static assert.

include/boost/unordered/detail/foa.hpp
```diff
@@ -1163,11 +1171,20 @@ constexpr static float const mlf = 0.875f;
...
- *   - TypePolicy::move(element_type&) returns a temporary object for value
- *     transfer on rehashing, move copy/assignment, and merge. For flat map, this
- *     object is a std::pair<Key&&,T&&>, which is generally cheaper to move
- *     than std::pair<const Key,T>&& because of the constness in Key. For
- *     node-based tables, this is used to transfer ownership of pointer.
+ *   - TypePolicy::construct and TypePolicy::destroy are used for the
+ *     construction and destruction of the internal types: value_type, init_type
+ *     and element_type. For flat containers, these are often all synonyms for
+ *     each other but for the node-based map, each one is a distinct type. These
+ *     are used for allocator-aware construction and destruction of the types
```
joaquintides, line 1178: It is not the case that for flat containers `value_type`, `init_type` and `element_type` are the same. In fact:
* For set containers, `init_type` and `value_type` are the same, for map containers they're not.
* For flat containers, `value_type` and `element_type` are the same, for node containers they're not,

Anyway, I'd just remove the "For flat containers [...] during insertion" part altogether, as it's redundant with the rest of the information already provided elsewhere in the comment.

joaquintides:
Comparison between latest n-running benchmark for `fix/uses_allocator` and baseline:

https://github.com/boostorg/boost_unordered_benchmarks/compare/f2e08733074a30b7e26bfe5e0dd5e5653d836464...a031e1b61c5f18532d56ff14814a4b3d87ed7660

joaquintides (approved):

pdimov:
What are we doing with this PR now that the release is closed for major changes?

joaquintides:
This is a bug fix, we're not allocator aware without it.

joaquintides:
As for time, we have it: I can merge as soon as Drone CI passes, and then we have more than one day to let the develop CI pass and merge to master.

pdimov:
All right then.

PR #186, opened 2023-02-28
Update changelog to reflect planned C++03 deprecation

cmazakas:
As voted on, 1.84.0 will no longer support C++03.


pdimov (approved):

joaquintides (commented):
Does this `{cpp}` macro really work? I've always used plain `C++` and it didn't seem to clutter the syntax.

PR #188, opened 2023-05-03
Feature/cfoa

joaquintides:


include/boost/unordered/detail/foa/rw_spinlock.hpp
```diff
@@ -0,0 +1,188 @@
+#ifndef BOOST_UNORDERED_DETAIL_FOA_RW_SPINLOCK_HPP_INCLUDED
+#define BOOST_UNORDERED_DETAIL_FOA_RW_SPINLOCK_HPP_INCLUDED
+
+// Copyright 2023 Peter Dimov
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+
+#include <boost/smart_ptr/detail/sp_thread_pause.hpp>
+#include <boost/smart_ptr/detail/sp_thread_sleep.hpp>
```
hadrielk, line 9: what a weird place to have those functions - I can see why no one wanted to put them in `Boost.Thread`, and `Boss.Atomic` isn't right either - but still it's... weird... to be in `SmartPtr`.

pdimov: They are part of the spinlock implementation in SmartPtr (which is itself an implementation detail.)

cmazakas:
I updated the CI files to not run on branch push. We should revert this before the official merge once we decided this candidate is good to go.

include/boost/unordered/concurrent_flat_map.hpp
```diff
@@ -0,0 +1,869 @@
...
+      {
+        table_ = ilist;
+        return *this;
+      }
+
+      /// Capacity
+      ///
+
+      size_type size() const noexcept { return table_.size(); }
+
```
joaquintides, line 305: `max_size` is missing.

include/boost/unordered/concurrent_flat_map.hpp
```diff
@@ -0,0 +1,869 @@
...
+      noexcept(noexcept(x.swap(y)))
+    {
+      x.swap(y);
+    }
+
+    template <class K, class T, class H, class P, class A, class Predicate>
+    typename concurrent_flat_map<K, T, H, P, A>::size_type erase_if(
+      concurrent_flat_map<K, T, H, P, A>& c, Predicate pred)
+    {
+      return c.erase_if(pred);
```
joaquintides, line 762: Calling `concurrent_flat:map::erase_if` here causes `pred` to be copied, and it may be a noncopyable type. You can replace with `table.erase_if` as you're already doing [here](https://github.com/boostorg/unordered/blob/feature/cfoa/include/boost/unordered/unordered_flat_map.hpp#L735), for which you'll need to make golbal `erase_if`a friend of `concurrent_flat_map`  (again, as in `unordered_flat_map`).

joaquintides (commented):
Looks very good. Things I have **not** reviewed:
* Tests. They're quite dauting to me, but I've seen Chris progress through testing and feel confident he's tapped into all there is to exercise.
* Docs, but these have been reviewed over several times.
* Code under folder `foa`, which is primarly mine. Would be great if @pdimov and/or @cmazakas could take a look there.

include/boost/unordered/detail/fwd.hpp
```diff
@@ -61,4 +61,90 @@ namespace boost {
...
+#define BOOST_UNORDERED_CXX11_CONSTRUCTION 0
+#elif !defined(BOOST_NO_CXX11_ALLOCATOR)
+#define BOOST_UNORDERED_CXX11_CONSTRUCTION 1
+#endif
+#endif
+
+#if !defined(BOOST_UNORDERED_CXX11_CONSTRUCTION)
+#define BOOST_UNORDERED_CXX11_CONSTRUCTION 0
+#endif
+
```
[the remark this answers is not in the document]
cmazakas: The simple answer is to make the test suite work.

The test suite used internally is a tad leaky in that it relies on these macros being defined. Previously, this was always true because the macros were used internally by the implementation and they leaked out.

In this case, we simply define a dummy `fwd.hpp` header file so that the normal test utilities we already use like `UNORDERED_TEST` compile and run with minimal changes required.

There's a bit of messy history here. This `fwd.hpp` header has always existed and its macros always relied upon by the test runner. There were more macros in implementation.hpp that were also required. I solved this for foa by just including `implementation.hpp` directly as it defined a lot of macros used by the tests that existed for fca.

But for this new concurrent container, I wanted to avoid relying on the existing implementation.hpp so this was the bare minimum that needed to be relocated from implementation.hpp to make the tests work.

I'm happy you brought this up because I realized that the foa fwd headers were erroneously including this fwd.hpp so I've now since relocated those includes.

Issue #190, opened 2023-05-30
Minor extra-semicolon warning issued for clang builds


hadrielk:
This is a minor nit, but the file `unordered/detail/foa.hpp` has a couple functions with extra semi-colons that trigger `clang`'s `extra-semi` warning:

1. `maybe_caused_overflow()` on line `233`.
2. `~destroy_on_exit()` on line `1719`.

cmazakas:
It's easy to add this check to our CI.

For clarity, you're using this compiler option, correct?
```
-Wextra-semi-stmt
```

PR #194, opened 2023-06-21
Remove extraneous semicolons

cmazakas:
This should fix https://github.com/boostorg/unordered/issues/190

Note: I didn't add `-Wextra-semi -Wextra-semi-stmt` to the testing Jamfile because Boost.Move has a single extraneous `;` as well and because we use `-Werror`, this breaks _our_ build until Move is fixed.

Because we're going to drop Move as a dependency in the next release, we can add `-Wextra-semi -Wextra-semi-stmt` to the test Jamfile then.

This PR was accomplished by manually hand-editing Boost.Move and then running the test suite locally with:
```
<toolset>clang-14:<cxxflags>"-Wunused-template -Wextra-semi -Wextra-semi-stmt"
```

### DIFF
--- a/include/boost/unordered/concurrent_flat_map.hpp
+++ b/include/boost/unordered/concurrent_flat_map.hpp
@@ -662,7 +662,7 @@ namespace boost {
       float max_load_factor() const noexcept
       {
         return table_.max_load_factor();
-      };
+      }
       void max_load_factor(float) {}
       size_type max_load() const noexcept { return table_.max_load(); }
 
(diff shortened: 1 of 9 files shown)


pdimov:
Do please submit a PR against Boost.Move with the necessary changes.

pdimov (approved):

joaquintides:
I guess we should wait for that pending PR to Boost.Move to be approved so that we can add `-Wextra-semi -Wextra-semi-stmt` to our own tests before approving this.

joaquintides (approved):

PR #198, opened 2023-07-23
Feature/concurrent unordered interop

joaquintides:
PR in progress (testing not done yet, for instance), I'm opening it in draft mode for discussion

### DIFF
--- a/doc/unordered/changes.adoc
+++ b/doc/unordered/changes.adoc
@@ -10,6 +10,8 @@
 
 * Added `[c]visit_while` operations to `boost::concurrent_map`,
 with serial and parallel variants.
+* Added efficient move construction of `boost::unordered_flat_map` from
+`boost::concurrent_flat_map` and vice versa.
 
 == Release 1.83.0 - Major update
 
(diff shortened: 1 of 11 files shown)


pdimov:
Looks OK to me so far, but needs tests.

joaquintides:
> Looks OK to me so far, but needs tests.

Yes, I'd like to ask @cmazakas to fill that in. A couple of points:

- We don't need specialized assignment operators: `unordered_m=std::move(concurrent_m)` does the right thing via an intermediate `boost::unordered_flat_map` move constructtion.
- We discussed this some weeks ago: the unordered-->concurrent direction incurs an allocation for the access structure and is technically O(n), though expected to be very fast (zero initialization of spinlocks and insertion counters). My opinion is that this is OK as is: I don't think swtching between one type of container and the other is something the program will do many times in a hot loop or something. But I'd like to know your opinion, of course.

pdimov:
Speed is not the problem - the possibility to throw is. But since this is a heterogeneous move constructor, it should be fine.

cmazakas:
Everything looks good to me as well. Took me awhile to finally understand everything.

I've added some tests that test construction from unordered->concurrent and then vice versa.

This is an allocating move constructor, sure, but I don't think it'll be a problem in practice. I think on some level it's to be expected by users but only time will tell.

pdimov:
You should add move assignment tests as well.

pdimov (approved):

PR #199, opened 2023-07-24
Feature/visit until

joaquintides:
Tests pending from @cmazakas, PR open for discussion. I think the API is quite uncontroversial, note however `[c]visit_(until|while)` returns a `bool`,  whereas `[c]visit_all` returns `size_type` (non-parallel versions) or `void` (parallel versions).

### DIFF
--- a/doc/unordered/changes.adoc
+++ b/doc/unordered/changes.adoc
@@ -6,6 +6,11 @@
 :github-pr-url: https://github.com/boostorg/unordered/pull
 :cpp: C++
 
+== Release 1.84.0
+
+* Added `[c]visit_until` and `[c]visit_while` operations to `boost::concurrent_map`,
+with serial and parallel variants.
+
 == Release 1.83.0 - Major update
 
 * Added `boost::concurrent_flat_map`, a fast, thread-safe hashmap based on open addressing.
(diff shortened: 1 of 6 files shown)


pdimov:
Why do we need both _while and _until?

joaquintides:
Not more than the standard library needs `std::all_of`, `std::any_of` and `std::none_of`.  Besides, I couln't decided which is more "natural" (until, maybe).

pdimov:
I don't think the duplication helps anyone; it just makes one wonder what's the subtle difference between the two (hence my question - I wasn't sure whether they are duplicates or not).

_while is more natural in C++ because there's no `until` keyword.

joaquintides:
Dropped. Note that the remaining example:
```cpp
// finds the key to a given (unique) value

int  key = 0;
int  value = ...;
bool found = !m.visit_while([&](const auto& x) {
  if(x.second == value) {
    key = x.first;
    return false; // finish
  }
  else {
    return true;  // keep on visiting
  }
});

if(found) { ... }
```

is more contrived with `_while`, but I felt looking for a specific value is more interesting than the other, deleted snippet:

```cpp
// check if all values are != 0
bool all_ok = m.visit_while([&](const auto& x) {
  return x.second != 0;
});
```

If you guys think otherwise I can change this bit, of course.

pdimov (approved):

PR #200, opened 2023-07-24
Feature/detect reentrancy

joaquintides:


joaquintides:
PR open for discussion:

* Tests are already added.
* There's a rather extensive protocol in place for enabling/disabling/configuring this checking.
  * Maybe overkill?
  * Should we document it? (I'd rather not, much as we haven't for `BOOST_UNORDERED_(DISABLE|ENABLE)_SSE2`).
* I took a look at `__attribute__((capability("mutex")))` and am not sure how the thing is supposed to be used and whether we have to conditionally add it or not.

pdimov:
I don't think you should duplicate all the BOOST_ASSERT functionality. That's what it's for.

pdimov:
For the attribute, see https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#mutex-h, the `Mutex` class.

joaquintides:
> I don't think you should duplicate all the BOOST_ASSERT functionality. That's what it's for.

Do we at least keep `BOOST_UNORDERED_DISABLE_REENTRANCY_CHECK`, or not even that?

pdimov:
Yes, we should have a way to disable the reentrancy check independently, for performance reasons.

cmazakas:
This looks good to me as well.

I feel like we should document the `BOOST_UNORDERED_DISABLE_REENTRANCY_CHECK` macro, no? It seems like it's automatically disabled in release builds when `NDEBUG` is defined but I think it'd be good if users know about this configuration macro in particular.

joaquintides:
Regarding support of Clang thread safety analysis:

I've written a draft of what it'd take to support it [here](https://github.com/boostorg/unordered/commit/f85ec3e9fb02b82cff1bcd5b4ce7789efbef0866). I've found severe limitations:

**Reentrancy detection** 
The anaysis is strictly [intra-procedural](https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#no-inlining). So, reentrancy is not detected here:

```cpp
using mutex=boost::unordered::detail::foa::annotated_mutex<boost::unordered::detail::foa::rw_spinlock>;
using lock_guard=boost::unordered::detail::foa::reentrancy_checked<boost::unordered::detail::foa::lock_guard<mutex>>;

static mutex m;

void foo()
{
  lock_guard lck{nullptr,m};
}

void bar()
{
  lock_guard lck{nullptr,m};
  foo();
}

int main()
{
  bar();
}
```
and the following explicit annotation is needed:
```cpp
void foo() BOOST_UNORDERED_EXCLUDES(m)
{
  lock_guard lck{nullptr,m};
}

void bar()
{
  lock_guard lck{nullptr,m};
  foo(); // warning : cannot call function 'foo' while mutex 'm' is held [-Wthread-safety-analysis]
}
```
which is a PITA as we'd need to annotate all the internal and external functions of `foa::concurrent_table`. Even so, the following reentrancy is again not detected:
```cpp
void foo() BOOST_UNORDERED_EXCLUDES(m)
{
  lock_guard lck{nullptr,m};
}

template<typename F>
void bar(F f)
{
  lock_guard lck{nullptr,m};
  f();
}

int main()
{
  bar([]{foo();});
}
```
because the lambda function is not annotated --and can't be, being user code. So, this Clang's analysis is useless for detecting reentrancies in user code.

**Race detection**

To detect unprotected access to data, the analysys tool requires annotations like this:
```cpp
static mutex m;
static int balance BOOST_UNORDERED_GUARDED_BY(m);
```
We can't have this annotation in our data structure:
* Container level: [protected data](https://github.com/boostorg/unordered/blob/bd24dfd284dbc70e7521915af0d8d049f74a1e85/include/boost/unordered/detail/foa/core.hpp#L1786-L1787) lives inside base class `foa::table_core`, which can't access the mutex in the derived class (except with incredible contortions).
* Group level: there is no way to tell the analysis tool that the _n_-th group in the `group` array is guarded by the _n_-th mutex in the `group_access` array: annotations can only use compile time expressions (typically, simple references to data members).

For all these reasons, I think we should abandon this line.

pdimov:
> For all these reasons, I think we should abandon this line.

No objections from me.

joaquintides:
> I feel like we should document the BOOST_UNORDERED_DISABLE_REENTRANCY_CHECK macro, no? It seems like it's automatically disabled in release builds when NDEBUG is defined but I think it'd be good if users know about this configuration macro in particular.

I'd rather leave this undocumented (and so little publicized), just as we did with `BOOST_UNORDERED_DISABLE_SSE2` etc. I don't think there'll be many situations where the user may legitimately need to disable it (`entry_trace` lists will typically be <=2 in size), and there are situations where it can be unlegitimately disabled to let code pass that we have officially banned.

pdimov:
I agree that `BOOST_UNORDERED_DISABLE_REENTRANCY_CHECK` should be documented. It's not an implementation detail or a testing artifact; its purpose is to be user-settable, and it has a specific and legitimate use - to disable reentrancy checking in debug builds for performance reasons.

joaquintides:
Well, I begrudginly concede :-) There's two places where this can be documented:

* In the [tutorial](https://200.unordered.prtest2.cppalliance.org/libs/unordered/doc/html/unordered.html#concurrent_visitation_based_api), which is less official.
* [Here](https://200.unordered.prtest2.cppalliance.org/libs/unordered/doc/html/unordered.html#concurrent_flat_map_concurrency_requirements_and_guarantees), as an implementation note (more official).

Opinions?

pdimov:
My suggestion is to either add a "Configuration Macros" subsection to the reference (after the "Guarantees") and put it there, or add a dedicated "Reentrancy Detection" section.

pdimov (approved):

PR #207, opened 2023-08-17
Feature/serialization support

joaquintides:
If we're OK with this PR, we need to sync up its merging with the removal of serialization functionality in

https://github.com/boostorg/serialization/blob/boost-1.83.0/include/boost/serialization/boost_unordered_map.hpp
https://github.com/boostorg/serialization/blob/boost-1.83.0/include/boost/serialization/boost_unordered_set.hpp

I can contact Robert for this as soon as the PR is approved. Thank you!


test/objects/test.hpp
```diff
@@ -1,6 +1,7 @@
 
 // Copyright 2006-2009 Daniel James.
 // Copyright 2022 Christian Mazakas
+// Copyright 2023 Joaqui M Lopez Munoz
```
cmazakas, line 4: I think we're missing an `n` here.

include/boost/unordered/detail/archive_constructed.hpp
```diff
@@ -0,0 +1,71 @@
...
+{
+  template<class Archive>
+  archive_constructed(const char* name,Archive& ar,unsigned int version)
+  {
+    core::load_construct_data_adl(ar,&get(),version);
+    BOOST_TRY{
+      ar>>core::make_nvp(name,get());
+    }
+    BOOST_CATCH(...){
+      (&get())->~T();
```
cmazakas, line 36: Out of curiosity, is there a reason to use `(&get())->~T();` over something like `get().~T();`?

joaquintides: No reason, I copied this code from some othe rlibrary of mine. Simplified as suggested, plus s/`&`/`boost::addressof` in the remaining cases.

test/objects/test.hpp
joaquintides: Fixed!

PR #213, opened 2023-09-18
self-include tests

cmazakas:
Finally add a missing piece of the test suite

Ensure all project headers compile with:
```cpp
#include <boost/unordered/...>
int main() {}
```


joaquintides:
Maybe, instead of having as many different test executables as there are headers, we can have multiple obj files linked into one executable? This probably takes less CI time.

cmazakas:
> Maybe, instead of having as many different test executables as there are headers, we can have multiple obj files linked into one executable? This probably takes less CI time.

Typically I prefer anything that enables me to run just a single test, should I ever need to, but I can look into this.

PR #217, opened 2023-10-09
Feature/bulk visit

joaquintides:
Some points for discussion:
* I opted for a very generic `visit(first, last, f)` API where `std::distance(first, last)` is not known at compile time: benchmarks indicate there's no significant performance penalty vs. a fixed-size API.
* I used plain `"visit"` rather than say `"bulk_visit"` for symmetry with other iterator operations like `insert(first, last)`.
* The iterators passed have to be forward (at least). The way I've implemented it, we check the iterator category in C++11-C++17 and the concept `std::forward_iterator` in C++20 and later: these two checks are not exactly equivalent.
* If  `Hash`, `Pred`  are not transparent, `*first` must dereference _exactly_ to a (const) (reference to) `key_type`, instead of allowing for implicit conversions to `key_type`.  This is done out of conservatism cause `*first` is called twice for each key, and the conversion may be expensive.
* I'm not providing a ready-made buffering utility such as devised [here](https://github.com/boostorg/boost_unordered_benchmarks/blob/c3f2ad693b09406566c955243949703562788322/parallel_load.cpp#L153-L182): I think this is easy enough for the user to implement by themselves that we don't need to bother.
* `bulk_visit_size` is a constant member value of the container rather than a global constant, even though it's the same everywhere (16). I feel this allows for more flexibility looking forward, but the ergonomics are a little worse for users implementing their buffering utilities.
* Benchmarks look very good, except: in GCC 12 x86, including `gtl::parallel_flat_hash_map` in the benchmarks makes the performance of `boost::concurrent_flat_map` in bulk mode [degrade](https://github.com/boostorg/boost_unordered_benchmarks/commit/6c160eb6d2193c6aa3352130fe25c216413b42e7) (!!). This does not make sense at all, but such is life. If you have a clue please let me know.


pdimov:
It might be useful for user to know whether `unordered_x<...>::bulk_visit_size` depends on the template parameters of the container, but I'm not quite sure how to express that.

joaquintides:
Well, the fact that `bulk_visit_size` is a nested value of `concurrent_x<...>` implies that it depends on `...`.  If we want to communicate this is not the case, we just define it as a lobal constant ourside `concurrent_x`.

pdimov:
Does it have to depend on them, though? Or on all of them? How would a different hash function possibly affect it?

joaquintides:
Well, if we were to fine tune `bulk_visit_size`, it'd depend on `Hash` (the faster `Hash`, the larger `bulk_visit_size` ought to be) and on the CPU architecture (memory fetching time/instruction throughput). No other dependencies, really. The hot code is this:

https://github.com/boostorg/unordered/pull/217/files#diff-a299ea8b15c2c2c34afbb92c11f11dc383bb2764a341c3e145090b6df0170720R1207-R1221

Maybe all these considerations are moot, anyway, as we can always increase `bulk_visit_size` wholesale is there's a need: there should be no practical downside to having this constant larger than the minimum required to achieve no waiting cycles --more data stored on the stack, that's all.

PR #218, opened 2023-10-21
Fix/remove lifted ops

joaquintides:
Opening the PR for discussion. I've removed all lifting `using` directives for `swap` and `operator(==|!=)`, _except_ in the case of `boost::unordered_[multi](map|set)`, and I wonder if we should go ahead and remove those too --I can't think of any legitimate use case where they could be needed.

### DIFF
--- a/include/boost/unordered/concurrent_flat_map.hpp
+++ b/include/boost/unordered/concurrent_flat_map.hpp
@@ -855,8 +855,6 @@ namespace boost {
 #endif
 
   } // namespace unordered
-
-  using unordered::concurrent_flat_map;
 } // namespace boost
 
 #endif // BOOST_UNORDERED_CONCURRENT_FLAT_MAP_HPP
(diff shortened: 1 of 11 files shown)


pdimov:
We should remove the 'lifting' for `operator==` and `operator!=` because only test suites do `boost::operator==(x, y)` instead of the correct `x == y`. For `swap`, things are less clear because we had a function `boost::swap` in Core so it's quite possible for code to have done `boost::swap(x, y)` for `unordered_set` or `unordered_map`.

If we remove the using directive, this will probably stop compiling, or start doing who knows what.

joaquintides:
> If we remove the using directive, this will probably stop compiling, or start doing who knows what.

From what I can gather from [boost::swap reference](https://www.boost.org/doc/libs/1_83_0/libs/core/doc/html/core/swap.html), ADL should kick in, so it seems safe to me to drop the lift. I've tested the following locally (on the `fix/remove_lifted_ops` branch):

```cpp
#include <boost/unordered/unordered_flat_map.hpp>
#include <boost/core/swap.hpp>

int main()
{
#pragma warning(disable : 4996)

  boost::unordered_flat_map<int,int> fm1,fm2;
  boost::swap(fm1,fm2);
}
```

and everything works fine. (BTW, the docs for `boost::core` do not indicated the function is deprecated.)

pdimov:
It won't work if boost/core/swap.hpp isn't included, though.

pdimov (approved):

PR #220, opened 2023-11-07
Fix `operator->` for `local_iterator`s

vslashg:
Invoking this operator causes a compile break.  This appears to have been caused by b1a9cde69.

### DIFF
--- a/include/boost/unordered/detail/fca.hpp
+++ b/include/boost/unordered/detail/fca.hpp
@@ -317,8 +317,7 @@ namespace boost {
         grouped_local_bucket_iterator() : p() {}
 
         reference operator*() const noexcept { return dereference(); }
-
-        pointer operator->() const noexcept { return boost::to_address(p); }
+        pointer operator->() const noexcept { return get_address(); }
 
         grouped_local_bucket_iterator& operator++() noexcept
         {
@@ -354,6 +353,7 @@ namespace boost {
         grouped_local_bucket_iterator(node_pointer p_) : p(p_) {}
 
         value_type& dereference() const noexcept { return p->value(); }
+        value_type* get_address() const noexcept { return p->value_ptr(); }
 
         bool equal(const grouped_local_bucket_iterator& x) const noexcept
         {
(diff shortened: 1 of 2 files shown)


pdimov:
What kind of compile break does it cause?

vslashg:
`lit->first` fails to compile; `operator->` for local iterators attempts to return a pointer to the node rather than a pointer to a value, so any attempt to instantiate this method (e.g by using `->` to dereference a `local_iterator`) fails to compile due to an impossible pointer conversion.

The PR contains a new regression test that will demonstrate the problem.

joaquintides:
Hi @vslashg, we've confirmed the bug you've spotted. The fix will arrive in Boost 1.84 (but after Boost 1.84, beta, which is due today). Thank you!

cmazakas:
Thank you for updating the tests!

The fix seems more complicated than it needs to be.

Why not just use:
```cpp
pointer operator->() const noexcept
{
  return std::addressof(dereference());
}
```
?

Also, we need to add coverage for `.cbegin()` so we test the `const_local_iterator` arm as well. It's fine to only update the unordered_map_test here but I'd like to see some coverage for the `unordered_set_test` as well.

Personally, what I'd like to do here is update the `bucket_tests.cpp` file. You don't necessarily have to do a whole new test case, you can just update the existing one there. But I'd really like to see some tests that test more than just "compile fine" which is what the compile_tests are for.

I can get this done but I wanted to offer the chance to iterate on the PR.

vslashg:
I'm certainly willing to make the changes you suggest, but for such a small change it may make more sense for you to implement the fix.  Let me know what you prefer; I'm happy to see this fixed either way.

cmazakas:
> but for such a small change it may make more sense for you to implement the fix.

Sure. This is an opportunity to become a contributor though! I wouldn't pass it up.

For the tests, I think we'd cover our bases if we just proved that `local_it.operator->()` yields the same address as `*local_it`.  The test suite already tests dereferencing the local iterators in other places like `invariants.hpp` so we don't need to go overboard here, just that `operator->()` is well-formed and returns the correct pointer value.

vslashg:
The friction of coming up with a test object such that `set_local_it->member` was well-formed was what discouraged me from writing a set test.  If you're happy with the test using explicit syntax of `set_local_it.operator->()` as a test invocation then things are much simpler.

I should be able to get to this either today or tomorrow.  Thanks!

joaquintides:
Problem fixed in #221.

PR #227, opened 2023-12-27
Optimize `emplace()` for exactly a `value_type` or `init_type` argument

### LINKED ISSUE #225: Optimize `emplace()` for exactly a `value_type` or `init_type` argument
Creating an intermediate value can be bypassed since we're already given one.

k3DW:
Closes #225

### DIFF
--- a/test/cfoa/emplace_tests.cpp
+++ b/test/cfoa/emplace_tests.cpp
@@ -170,6 +170,74 @@ namespace {
     }
   } lvalue_emplace_or_visit;
 
+  struct copy_emplacer_type
+  {
+    template <class T, class X> void operator()(std::vector<T>& values, X& x)
+    {
+      static constexpr auto value_type_cardinality =
+        value_cardinality<typename X::value_type>::value;
+
+      std::atomic<std::uint64_t> num_inserts{0};
+      thread_runner(values, [&x, &num_inserts](boost::span<T> s) {
+        for (auto const& r : s) {
+          bool b = x.emplace(r);
+          if (b) {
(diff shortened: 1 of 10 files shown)


test/cfoa/emplace_tests.cpp
```diff
@@ -170,6 +170,67 @@ namespace {
...
+            ++num_inserts;
+          }
+        }
+      });
+      BOOST_TEST_EQ(num_inserts, x.size());
+      BOOST_TEST_EQ(raii::default_constructor, 0u);
+
+      if (std::is_same<T, typename X::value_type>::value &&
+          !std::is_same<typename X::key_type,
+            typename X::value_type>::value) { // map value_type can only be
```
joaquintides, line 221: This generates a C4127 warning in VS. See [this](https://github.com/boostorg/unordered/blob/0d2751c5e1a99b1275717e26c16eb672150da77b/include/boost/unordered/detail/foa/core.hpp#L1493-L1496) and [this](https://github.com/boostorg/unordered/blob/0d2751c5e1a99b1275717e26c16eb672150da77b/include/boost/unordered/detail/foa/core.hpp#L1556-L1558) for a workaround.

k3DW: I tried to use `#pragma warning(suppress : 4127)` but that didn't seem to work. I'll go with the `push`, `disable`, `pop` method.

k3DW:
I made a mistake and forgot to add the new `emplace_smf_tests.cpp` to the Jamfile

PR #230, opened 2024-01-20
Optimize `emplace()` for arguments of the form `k, v`

### LINKED ISSUE #226: Optimize `emplace()` for arguments of the form `k, v`
We can defer the `value_type` construction until we're sure we will emplace.

k3DW:
Closes #226

### DIFF
--- a/doc/unordered/changes.adoc
+++ b/doc/unordered/changes.adoc
@@ -6,6 +6,11 @@
 :github-pr-url: https://github.com/boostorg/unordered/pull
 :cpp: C++
 
+== Release 1.85.0
+
+* Optimized `emplace()` for a `value_type` or `init_type` (if applicable) argument to bypass creating an intermediate object. The argument is already the same type as the would-be intermediate object.
+* Optimized `emplace()` for `k,v` arguments on map containers to delay constructing the object until it is certain that an element should be inserted. This optimization happens when the map's `key_type` is move constructible or when the `k` argument is a `key_type`.
+
 == Release 1.84.0 - Major update
 
 * Added `boost::concurrent_flat_set`.
--- a/include/boost/unordered/detail/allocator_constructed.hpp
+++ b/include/boost/unordered/detail/allocator_constructed.hpp
@@ -1,0 +1,59 @@
+/* Copyright 2024 Braden Ganetsky.
+ * Distributed under the Boost Software License, Version 1.0.
+ * (See accompanying file LICENSE_1_0.txt or copy at
+ * http://www.boost.org/LICENSE_1_0.txt)
+ *
+ * See https://www.boost.org/libs/unordered for library home page.
+ */
+
+#ifndef BOOST_UNORDERED_DETAIL_ALLOCATOR_CONSTRUCTED_HPP
+#define BOOST_UNORDERED_DETAIL_ALLOCATOR_CONSTRUCTED_HPP
+
+#include <boost/core/allocator_traits.hpp>
+#include <boost/unordered/detail/opt_storage.hpp>
+
+namespace boost {
+  namespace unordered {
+    namespace detail {
+
+      struct allocator_policy
+      {
+        template <class Allocator, class T, class... Args>
+        static void construct(Allocator& a, T* p, Args&&... args)
+        {
+          boost::allocator_construct(a, p, std::forward<Args>(args)...);
+        }
+
+        template <class Allocator, class T>
+        static void destroy(Allocator& a, T* p)
+        {
+          boost::allocator_destroy(a, p);
+        }
+      };
+
+      /* constructs a stack-based object with the given policy and allocator */
+      template <class Allocator, class T, class Policy = allocator_policy>
+      class allocator_constructed
+      {
+        opt_storage<T> storage;
+        Allocator alloc;
+
+      public:
+        template <class... Args>
+        allocator_constructed(Allocator const& alloc_, Args&&... args)
+            : alloc(alloc_)
+        {
+          Policy::construct(
+            alloc, storage.address(), std::forward<Args>(args)...);
+        }
+
+        ~allocator_constructed() { Policy::destroy(alloc, storage.address()); }
+
+        T& value() { return *storage.address(); }
+      };
+
+    } /* namespace detail */
+  }   /* namespace unordered */
+} /* namespace boost */
+
+#endif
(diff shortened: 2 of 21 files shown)


include/boost/unordered/detail/implementation.hpp
```diff
@@ -1913,6 +1917,44 @@ namespace boost {
...
         }
 
+      private:
+        class allocator_constructed_key_type
+        {
+          using key_type = typename Types::key_type;
+          using key_allocator_type =
+            typename detail::rebind_wrap<node_allocator_type, key_type>::type;
+
+          alignas(key_type) unsigned char storage[sizeof(key_type)];
```
cmazakas, line 1928: We already have this pattern in the codebase, check out https://github.com/boostorg/unordered/blob/67c5cdb3a69f0b92d2779880ce9aa1d46e54cf7b/include/boost/unordered/detail/opt_storage.hpp

include/boost/unordered/detail/implementation.hpp
```diff
@@ -1913,6 +1917,44 @@ namespace boost {
           }
         }
 
+      private:
+        class allocator_constructed_key_type
+        {
+          using key_type = typename Types::key_type;
+          using key_allocator_type =
+            typename detail::rebind_wrap<node_allocator_type, key_type>::type;
```
cmazakas, line 1926: `construct()` is also a template so you don't need to rebind here

doc/unordered/changes.adoc
```diff
@@ -6,6 +6,11 @@
 :github-pr-url: https://github.com/boostorg/unordered/pull
 :cpp: C++
 
+== Release 1.85.0
+
+* Optimized `emplace()` for a `value_type` or `init_type` (if applicable) argument to only allocate once.
```
cmazakas, line 11: The use of "allocate once" is confusing to me here. What's meant by it?

include/boost/unordered/detail/foa/core.hpp
```diff
@@ -1350,6 +1407,19 @@ table_core:empty_value<Hash,0>,empty_value<Pred,1>,empty_value<Allocator,2>
...
+  using is_map =
+    std::integral_constant<bool, !std::is_same<key_type, value_type>::value>;
+  using is_key_type_move_constructible =
+    typename std::conditional<detail::is_complete<key_type>::value,
+      std::is_move_constructible<key_type>, std::false_type>::type;
+  template <typename K>
+  using is_emplace_kv_able = std::integral_constant<bool,
+    is_map::value && (detail::is_similar<K, key_type>::value ||
+                       is_key_type_move_constructible::value)>;
+
```
cmazakas, line 1421: Is there a way to keep these traits from having to be protected member typedefs?

include/boost/unordered/detail/implementation.hpp
k3DW: I tried using `opt_storage` here and it's fine, but it gives a compile error because of deleted copy constructor in the `detail::foa::allocator_constructed` class created for this exact same purpose. I think instead I'll factor these 2 classes into a single implementation

doc/unordered/changes.adoc
k3DW: I'll make it more clear. When we're given a `value_type` or `init_type` as an argument to `emplace()`, we can bypass creating an intermediate value, which we would otherwise create.

include/boost/unordered/detail/foa/core.hpp
k3DW: They can be free traits instead. I'm not sure about the naming though. I'll take any suggestions

cmazakas (approved):

include/boost/unordered/detail/map.hpp
```diff
@@ -18,6 +19,7 @@ namespace boost {
         typedef std::pair<K const, M> value_type;
         typedef H hasher;
         typedef P key_equal;
+        typedef K key_type;
         typedef K const const_key_type;
```
joaquintides, line 23: Why has you added this `typedef`? Seems obvious it was missing, but is there something that used to fail without it?

doc/unordered/changes.adoc
```diff
@@ -6,6 +6,11 @@
 :github-pr-url: https://github.com/boostorg/unordered/pull
 :cpp: C++
 
+== Release 1.85.0
+
+* Optimized `emplace()` for a `value_type` or `init_type` (if applicable) argument to bypass creating an intermediate object. The argument is already the same type as the would-be intermediate object.
+* Optimized `emplace()` for `k,v` arguments on map containers to delay constructing the object until it is certain that an element should be inserted. This optimization when the map's `key_type` is move constructible or when the `k` argument is a `key_type`.
+
```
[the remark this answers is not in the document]
k3DW: I missed the word "occurs", thanks

test/unordered/init_type_insert_tests.cpp
```diff
@@ -12,12 +12,6 @@
 
 #include <boost/config.hpp>
 
-#if defined(BOOST_LIBSTDCXX_VERSION)
-#if BOOST_LIBSTDCXX_VERSION < 60000
-#define BOOST_UNORDERED_NO_INIT_TYPE_TESTS
-#endif
-#endif
-
```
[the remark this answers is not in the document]
k3DW: I didn't remove this, I just put this macro definition into "test/helpers/unordered.hpp" so it can be used by multiple test files

include/boost/unordered/detail/allocator_constructed.hpp
```diff
@@ -0,0 +1,62 @@
...
+        template <class Allocator, class T>
+        static void destroy(Allocator& a, T* p)
+        {
+          boost::allocator_destroy(a, p);
+        }
+      };
+
+      /* constructs a stack-based object with the given allocator */
+      template <class Allocator, class T>
+      using allocator_constructed =
```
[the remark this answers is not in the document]
k3DW: I agree with this, I'll make the change

include/boost/unordered/detail/map.hpp
k3DW: Good catch, I used it in an earlier draft, but I forgot to remove it later. I'll remove it here, it's not needed

include/boost/unordered/detail/implementation.hpp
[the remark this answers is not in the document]
k3DW: Good catch, I'll add this test and fix the regression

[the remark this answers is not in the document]
k3DW: This would have been detected if I wrote tests for immovable fca maps too. I added those tests and fixed the problem

joaquintides (approved):

pdimov (approved):

cmazakas (approved):

PR #234, opened 2024-02-17
Fixed support for allocators with `explicit` copy constructors

### LINKED ISSUE #204: Unordered doesn't respect explicit converting Allocator constructors
Relevant Compiler Explorer link: https://godbolt.org/z/6dzPsTxob

joaquintides:
Fixes #204


pdimov:
Interesting. Are such allocators conforming? The latest allocator requirement table seems to say so; was it always such, or is this a recent change?

pdimov:
(Apart from those philosophical questions, I don't see anything wrong with the change.)

pdimov:
Seems intentional because the same type entries explicitly mandate `A a2 = a;` to also work.

pdimov:
Looks like C++03 only had the `X a(b);` requirement, C++11 added the `X a1(a);` requirement, and C++14 added `X a1 = a;`.

This seems a bit accidental to me.

joaquintides:
> This seems a bit accidental to me.

Probably. If I recall correctly, though, Chris commented this issue actually hit a real user, so it dones't hurt to have it fixed.

pdimov:
Yeah, I know, I was just wondering whether such allocators are allowed.

pdimov (approved):

PR #239, opened 2024-03-23
Add unordered container `pmr` aliases

### LINKED ISSUE #229: Add `pmr` aliases for the containers
For C++>=17, use `std::pmr`.

This should not be supported in C++<17.

k3DW:
Closes #229

### DIFF
--- a/test/cfoa/pmr_allocator_tests.cpp
+++ b/test/cfoa/pmr_allocator_tests.cpp
@@ -1,0 +1,8 @@
+// Copyright (C) 2024 Braden Ganetsky
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#define BOOST_UNORDERED_CFOA_TESTS
+#include <boost/unordered/concurrent_flat_map.hpp>
+#include <boost/unordered/concurrent_flat_set.hpp>
+#include "../unordered/pmr_allocator_tests.cpp"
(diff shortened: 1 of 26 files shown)


test/cfoa/pmr_allocator_tests.cpp
joaquintides, line 1: This is a near-perfect duplicate of `test/unordered/pmr_allocator_tests.cpp`. When cfoa was introduced, its tests were kept separate because the interface of concurrent containers is too dissimilar to that of regular containers, but in this case the only relevant difference in the testing code is [this portion](https://github.com/k3DW/unordered/blob/feature/229/test/cfoa/pmr_allocator_tests.cpp#L75-L91). Consider unifying the same way as it's done in [explicit_alloc_ctor_tests.cpp](https://github.com/k3DW/unordered/blob/feature/229/test/cfoa/explicit_alloc_ctor_tests.cpp)

test/unordered/pmr_allocator_tests.cpp
```diff
@@ -0,0 +1,219 @@
...
+  typename std::enable_if<!test::is_map<X>::value, std::size_t>::type
+  emplace_strings(X& x)
+  {
+    std::string_view sv =
+      "this is a string that's longer than the SBO threshold";
+    x.emplace(sv);
+    // Return how many chars were allocated using a pmr allocator
+    return std::is_same<typename X::key_type, std::pmr::string>::value
+             ? sv.size()
+             : 0;
```
joaquintides, line 77: Same as in `test/cfoa/pmr_allocator_tests.cpp`.

pdimov (approved):

Issue #162, opened 2022-11-14
Implement .natvis files for containers


cmazakas:
gdb pretty-printers will probably also be needed.

PR #249, opened 2024-05-28
Implement natvis visualizations

k3DW:
Partial work on #162. This applies to containers with raw pointers as the allocator pointer type. This does not account for fancy pointers.


k3DW:
Thanks for the comments. I'll address each one of them below.
* Non-SIMD layout *does* work for me, so we can sort that out in private. Typically when I switch, it requires an entire "rebuild" in Visual Studio, not just a "build".
* Yes this is exactly how I detect regular layout. See [here](https://github.com/boostorg/unordered/blob/feature/natvis/extra/boost_unordered.natvis#L405-L406) for example.
* "mo" just means "match occupied". It's a helper intrinsic for the actual `__match_occupied` intrinsic. It seems like this is too confusing, so I should use the full name instead of abbreviating.
* Likewise "issen" means "is sentinel". These are helpers for `__is_sentinel`. Similarly, it looks like I should use the full name here for clarity.
* Can do. I'll rename the `baseN` aliases and leave a comment explaining it's for natvis.
* The link you provided does not show `table_iterator::operator*()`. This shows dereferencing the member `table_iterator::p_`, which has the type `table_iterator::table_element_pointer`. Without fancy pointers, this is a regular pointer that can be dereferenced as a primitive operation. This is not calling any user-defined `operator*()`.
* I'll see what I can do about the visualization of `end()` iterators.
* I'll also work on the docs and release notes. No need to provide it. I'll take feedback on that when I do it.
* I'm not familiar with the stats, but we can talk about that as well.
* Supporting individual types of fancy pointers is possible, but I will make the case that support for generalized fancy pointers is impossible as a single unified implementation. As I mentioned a few points up, this is not calling any user-defined `operator*()`, which would be disallowed in natvis. I think more discussion of supporting specific types of fancy pointers (like `boost::interprocess::fancy_pointer` for example) should come after the initial natvis implementation here. Then we can reassess.

joaquintides:
> Non-SIMD layout does work for me, so we can sort that out in private. Typically when I switch, it requires an entire "rebuild" in Visual Studio, not just a "build".

I confirm it works for non-SIMD and/or x86 on my side after rebuilding. Thank you!

> "mo" just means "match occupied" [...] Likewise "issen" means "is sentinel".

OK, no need to change names if it works for you.

> The `baseN` aliases are following the same naming scheme as `boost::unordered::detail::compressed`, which has `base1` and `base2` aliases.

All right, makes sense. Please do add the comment, though.

> The link you provided does not show table_iterator::operator*() [...]

Thank you for the explanation. I stand corrected.

> Supporting individual types of fancy pointers is possible, but I will make the case that support for generalized fancy pointers is impossible as a single unified implementation[...]

Yes, you're right. I got sidetracked by my wrong assumption that code functions could indeed be invoked, which is not the case as discussed above. So, let's forget about fancy pointers for now.

One additional observation:

* I see you've written two versions of the element loop, one for regular layout and one for non-regular, basically replicating the code in `table_iterator::increment`.  Turns out that the version of `increment` for regular layout is there for performance reasons, and the non-regular-layout version should work equally well. So, you probably can retain only the non-regular-layout loop --the mo and issen functions, however, need be versioned for regular/non-regular.

k3DW:
> OK, no need to change names if it works for you.

I disagree. If there's a point of possible confusion, I want to remove it. I don't want to be the only one who can maintain this code. I'll change the names.

> So, you probably can retain only the non-regular-layout loop...

Huh you're right. That removed over 100 lines. Good catch.

extra/boost_unordered.natvis
```diff
@@ -0,0 +1,380 @@
...
+            <Item Name="[successful_lookup]">successful_lookup</Item>
+            <Item Name="[unsuccessful_lookup]">unsuccessful_lookup</Item>
+        </Expand>
+    </Type>
+
+    <Type Name="boost::unordered::detail::cumulative_stats&lt;*&gt;" Inheritable="true">
+        <Intrinsic Name="bit_cast_to_double" Expression="*reinterpret_cast&lt;double*&gt;(&amp;i)">
+            <Parameter Name="i" Type="size_t" />
+        </Intrinsic>
+        <Intrinsic Name="bit_cast_to_size_t" Expression="*reinterpret_cast&lt;size_t*&gt;(&amp;d)">
```
joaquintides, line 177: This crashes in x86 mode:

imagen

I could fix it locally by rewriting as follows:
```
        <Intrinsic Name="bit_cast_to_double" Expression="*reinterpret_cast&lt;double*&gt;(&amp;i)">
            <Parameter Name="i" Type="uint64_t" />
        </Intrinsic>
        <Intrinsic Name="bit_cast_to_uint64_t" Expression="*reinterpret_cast&lt;uint64_t*&gt;(&amp;d)">
            <Parameter Name="d" Type="double" />
        </Intrinsic>
        
        <!-- https://en.wikipedia.org/wiki/Fast_inverse_square_root#Magic_number -->
        <Intrinsic Name="__inv_sqrt_init" Expression="bit_cast_to_double(0x5FE6EB50C7B537A9ull - (bit_cast_to_uint64_t(x) &gt;&gt; 1))">
            <Parameter Name="x" Type="double" />
        </Intrinsic>
```

(FWIW, one of the changes (s/0x5FE6EB50C7B537A9/0x5FE6EB50C7B537A9ul) is probably not necessary, but felt it look cleaner.)

k3DW: Thanks, I'll make those changes.

joaquintides (approved):

PR #250, opened 2024-05-29
Feature/is_avalanching_trait_update

### LINKED ISSUE #248: Change `is_avalanching` protocol to one based on `std::[true|false]_type`
Currently, hash function avalanching is indicated by the mere presence of a nested `is_avalanching` typedef (like `is_transparent`). Change that to requiring that `is_avalanching` is  `std::[true|false]_type` (or derived from those), in the same vein as propagate traits for allocators.

To keep backwards compatiblity, if `is_avalanching` is present but not (derived from) `std::[true|false]_type` , this must be taken as equivalent to `using is_avalanching = std::true_type`.

joaquintides:
Fixes #248.

### DIFF
--- a/doc/unordered/hash_traits.adoc
+++ b/doc/unordered/hash_traits.adoc
@@ -36,9 +36,9 @@ this property is critical for the proper behavior of open-addressing hash contai
 
  * `false` if `Hash::is_avalanching` is not present,
  * `Hash::is_avalanching::value` if this is present and convertible at compile time to a `bool`,
- * `true` if `Hash::is_avalanching` is `void` (this usage is deprecated).
+ * `true` if `Hash::is_avalanching` is `void` (this usage is deprecated),
+ * ill-formed otherwise.
  
-The behavior is undefined if none of the three cases above is met.
 Users can then declare a hash function `Hash` as avalanching either by embedding an appropriate `is_avalanching` typedef
 into the definition of `Hash`, or directly by specializing `hash_is_avalanching<Hash>` to a class with
 an embedded compile-time constant `value` set to `true`.
--- a/doc/unordered/hash_traits.adoc
+++ b/doc/unordered/hash_traits.adoc
@@ -32,9 +32,14 @@ large changes in the returned hash code &#8212;ideally, flipping one bit in the 
 the input value results in each bit of the hash code flipping with probability 50%. Approaching
 this property is critical for the proper behavior of open-addressing hash containers.
 
-`hash_is_avalanching<Hash>::value` is `true` if `Hash::is_avalanching` is a valid type,
-and `false` otherwise.
-Users can then declare a hash function `Hash` as avalanching either by embedding an `is_avalanching` typedef
+`hash_is_avalanching<Hash>::value` is:
+
+ * `false` if `Hash::is_avalanching` is not present,
+ * `Hash::is_avalanching::value` if this is present and convertible at compile time to a `bool`,
+ * `true` if `Hash::is_avalanching` is `void` (this usage is deprecated).
+ 
+The behavior is undefined if none of the three cases above is met.
+Users can then declare a hash function `Hash` as avalanching either by embedding an appropriate `is_avalanching` typedef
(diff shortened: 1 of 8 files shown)


doc/unordered/hash_traits.adoc
```diff
@@ -32,9 +32,14 @@ large changes in the returned hash code &#8212;ideally, flipping one bit in the
...
-`hash_is_avalanching<Hash>::value` is `true` if `Hash::is_avalanching` is a valid type,
-and `false` otherwise.
-Users can then declare a hash function `Hash` as avalanching either by embedding an `is_avalanching` typedef
+`hash_is_avalanching<Hash>::value` is:
+
+ * `false` if `Hash::is_avalanching` is not present,
+ * `Hash::is_avalanching::value` if this is present and convertible at compile time to a `bool`,
+ * `true` if `Hash::is_avalanching` is `void` (this usage is deprecated).
+ 
+The behavior is undefined if none of the three cases above is met.
```
pdimov, line 41: I don't think we want "undefined" here.

doc/unordered/hash_traits.adoc
```diff
@@ -32,9 +32,14 @@ large changes in the returned hash code &#8212;ideally, flipping one bit in the
 the input value results in each bit of the hash code flipping with probability 50%. Approaching
 this property is critical for the proper behavior of open-addressing hash containers.
 
-`hash_is_avalanching<Hash>::value` is `true` if `Hash::is_avalanching` is a valid type,
-and `false` otherwise.
-Users can then declare a hash function `Hash` as avalanching either by embedding an `is_avalanching` typedef
+`hash_is_avalanching<Hash>::value` is:
+
+ * `false` if `Hash::is_avalanching` is not present,
```
pdimov, line 37: What do we want when `is_avalanching` is present, but is not a type? `false` or a compile error? The latter, probably.

doc/unordered/hash_traits.adoc
joaquintides: You're right, I miswrote that for "ill-formed". What's specifically ill formed? The "program", the expression `hash_is_avalanching<Hash>::value`, something else?

doc/unordered/hash_traits.adoc
joaquintides: We can add a specialization like this:
```cpp
template<typename Hash>
struct hash_is_avalanching_impl<
  Hash,
  typename std::enable_if<((void)Hash::is_avalanching,true)>::type
>{};
```

doc/unordered/hash_traits.adoc
pdimov: The program, I suppose.

doc/unordered/hash_traits.adoc
pdimov: Let's just document it as `false` and be done with it. It's way too convoluted otherwise.

doc/unordered/hash_traits.adoc
joaquintides: I think I'll go with "the expression", as the user can legitimately define and use something like:
```cpp
struct my_hash{ using is_avalanching=int; };
```
as long as he doesn't plug into Boost.Unordered.

doc/unordered/hash_traits.adoc
joaquintides: Maybe we _should_ make it a compile-time error to protect users from:
```cpp
struct my_hash
{
  static constexpr bool is_avalanching=true;
};
```

pdimov: That's exactly what I thought, but if it's too much of a hassle, we shouldn't go out of our way to try and catch that.

joaquintides: It's not a hassle, the change proposed above does indeed do the trick.

cmazakas: Fwiw, if Joaquin doesn't mind putting in the work, I also think making it a compiler error would be fantastic. As a user of Unordered, this is exacly the behavior I'd want (i.e. saving me from myself).

cmazakas (approved):

Issue #256, opened 2024-06-18
Using concurrent_flat_map with interprocess mapped memory allocator


bocharov:
Hello,

I have a use case that involves using Boost's concurrent_flat_map with an interprocess mapped memory allocator. My application is designed to be read-heavy with occasional inserts and deletes. It also requires multiple processes (not just threads) to access the container concurrently.

I have two questions:

1. Is it possible to use concurrent_flat_map with an interprocess mapped memory allocator in such a way that multiple processes can access and modify the map concurrently?
2. If the above is not feasible, could you recommend an alternative container or approach within Boost that would suit this use case? My primary goals are to minimize synchronization costs and ensure efficient read access across multiple processes.

Thank you for your assistance!

joaquintides:
Yes, it works, see this example:

https://github.com/boostorg/unordered/blob/develop/test/cfoa/interprocess_concurrency_tests.cpp

Note, however, that this feature is not documented, so you would be using it unofficially. We haven't decided yet whether to guarantee it for future releases of the library.

Feng-Zihao:
Hello, @joaquintides 

I think this feature is super great!  Strongly recommended to guarantee it for future release. At least for concurrent_flat_map.

PR #258, opened 2024-06-21
Feature/test interprocess concurrency

joaquintides:
Verifies that concurrent containers work when accessed through shared memory by different processes, including if stats are enabled (which case didn't work before).

Addresses #256.

### DIFF
--- a/extra/boost_unordered.natvis
+++ b/extra/boost_unordered.natvis
@@ -170,7 +170,7 @@ file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
         </Expand>
     </Type>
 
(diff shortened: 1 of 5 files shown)


cmazakas (approved):

cmazakas:
Hmm, this PR makes me wonder if we can dramatically simplify the shared memory names in the mmap_tests.

Looking at the previous test compared to this one, the old seems unideal in comparison.

pdimov (approved):

joaquintides:
> Hmm, this PR makes me wonder if we can dramatically simplify the shared memory names in the mmap_tests.

Yes, the difference is that the old code prepends a fixed string to the randomly generated UUID. I think this is superfluous as `boost::uuids::random_generator()()` provides the necessary diversity.

PR #262, opened 2024-07-12
Implement natvis for fancy pointers

k3DW:
Partial work on #162. This will come with a corresponding PR into boostorg/interprocess for the `offset_ptr` support, after merging this PR.

### DIFF
--- a/test/natvis_tests.cpp
+++ b/test/natvis_tests.cpp
@@ -1,0 +1,187 @@
+// Copyright 2024 Braden Ganetsky
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+
+// This test applies to MSVC only. This is a file for manual testing.
+// Run this test and break manually at the variable called `break_here`.
+// Inspect the variables using the Visual Studio debugger to test correctness.
+
+#if !defined(BOOST_MSVC)
+
+#include <boost/config/pragma_message.hpp>
+BOOST_PRAGMA_MESSAGE("These tests are for Visual Studio only.")
+int main() {}
+
+#else
+
+#if 0 // Change to `#if 1` to test turning off SIMD optimizations
+#define BOOST_UNORDERED_DISABLE_SSE2
+#define BOOST_UNORDERED_DISABLE_NEON
+#endif
+
+#define BOOST_UNORDERED_ENABLE_STATS
+
+#include <boost/core/lightweight_test.hpp>
+
+#include <boost/interprocess/allocators/allocator.hpp>
+#include <boost/interprocess/managed_shared_memory.hpp>
+#include <boost/unordered/concurrent_flat_map.hpp>
+#include <boost/unordered/concurrent_flat_set.hpp>
+#include <boost/unordered/unordered_flat_map.hpp>
+#include <boost/unordered/unordered_flat_set.hpp>
+#include <boost/unordered/unordered_map.hpp>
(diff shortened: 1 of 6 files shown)


test/natvis_tests.cpp
```diff
@@ -0,0 +1,185 @@
...
+template <class... Args> void use(Args&&...) {}
+
+using map_value_type = std::pair<const std::string, int>;
+using set_value_type = std::string;
+
+template <class Tester> void natvis_test(Tester& tester)
+{
+  // clang-format off
+    auto fca_map      = tester.template construct_map<boost::unordered_map>();
+    auto fca_multimap = tester.template construct_map<boost::unordered_multimap>();
```
cmazakas, line 45: Idiomatically, the test suite would pass `T* p = nullptr;` to these kinds of functions so the type could be deduced from the argument.

Not sure if this would wind up looking better or not, but maybe worth considering given how the test suite is mostly structured now.

test/natvis_tests.cpp
```diff
@@ -0,0 +1,185 @@
...
+#endif // defined(BOOST_MSVC)
+
+int main()
+{
+#if defined(BOOST_MSVC)
+
+  natvis_test(default_tester);
+  natvis_test(offset_ptr_tester);
+
+#endif // defined(BOOST_MSVC)
```
cmazakas, line 182: One thing to consider might be refactoring this to have only one `#ifdef BOOST_MSVC` and two alternate mains.

The shorter main can live at the top and have something like a `BOOST_PRAGMA_MESSAGE("skipping test for unsupported platform")`

test/natvis_tests.cpp
cmazakas, line 45: Fwiw, the way the test suite is structured now, Daniel James avoided a lot of  this kind of `template` stuff by declaring `T* p = nullptr` and then passing `p`, letting functions deduce the template param from the argument.

Not 100% sure if this would actually make the code nicer here but it's something to maybe consider. So, like:
```cpp
boost::unordered_map* fca_map = nullptr;
// ...

auto fca_map = tester.construct_map(fca_map);
// ...
```

I thought Daniel was pretty clever for this at the time when I started working on the codebase.

k3DW: I don't think that design works here for 2 reasons.

First, I can't create a pointer to the type without a bunch of extra mess. It would have to say something like `typename Tester::template map_type<boost::unordered_map>* fca_map = nullptr;`, which I think is uglier than just passing the class template name into `construct` directly.

Secondly, because of ownership and lifetime, the 2 different `Tester` types return different kinds of pointers. For `offset_ptr_tester`, the tester itself owns the memory, so it returns a raw pointer from the `construct` functions. But for `default_tester`, it needs to return a `unique_ptr` to satisfy the same API and not leak memory. Maybe I could write something like:
```cpp
typename Tester::template map_pointer_type<boost::unordered_map> fca_map = nullptr;
tester.construct_map(fca_map);
```
but I also think this is uglier than what I have already
```cpp
auto fca_map = tester.template construct_map<boost::unordered_map>();
```

I don't see any other way to make it look nicer, given that I want the code to be generic across different types of `Tester`. We may want to add another type of fancy pointer into these tests later on, and I want to make this easily extendable - just create another tester type, and the `natvis_test()` function stays exactly the same.

joaquintides (approved):
No further comments on my side.

cmazakas (approved):

PR #266, opened 2024-07-20
Add support for modular build structure.

grafikrobot:
This is part of the effort to make the Boost libraries "modular" for build and consumption. See https://lists.boost.org/Archives/boost/2024/01/255704.php and https://github.com/grafikrobot/boost-b2-modular/blob/b2-modular/README.adoc for more information.

This PR depends on the following other PRs being merged to both develop and master branches of the respective repos:

- https://github.com/boostorg/boost/pull/854

This PR will be changed to ready for review, i.e. not draft, when the above are merged. Do not merge this one until that time.


build.jam
```diff
@@ -0,0 +1,26 @@
+# Copyright René Ferdinand Rivera Morell 2023
```
k3DW, line 1: I assume this should say `2024` instead of `2023`

grafikrobot: No entirely incorrect... Some of the changes have been around since last year. But, yes, I should update that.

PR #271, opened 2024-08-05
Feature/concurrent node containers

joaquintides:
Adds `boost::concurrent_node_(map|set)`. The following is _not_ part of this PR and will be done later:

* Upadate concurrent benchmarks to include the new node-based containers.
* Include `(insert|emplace)_and_visit` operations. This is still under discussion (conversation on the #boost-unordered Slack group).


include/boost/unordered/detail/foa/core.hpp
```diff
@@ -1458,6 +1458,11 @@ table_core:empty_value<Hash,0>,empty_value<Pred,1>,empty_value<Allocator,2>
   using stats=table_core_stats;
 #endif
 
+#if defined(BOOST_GCC)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
cmazakas, line 1463: A `maybe-uninitialized` warning? What did the warning say, out of curiosity?

Historically, this might mean we need a launder() call somewhere.

extra/boost_unordered.natvis
```diff
@@ -421,6 +421,8 @@ file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
         <AlternativeType Name="boost::unordered::unordered_node_set&lt;*&gt;" />
         <AlternativeType Name="boost::unordered::concurrent_flat_map&lt;*&gt;" />
         <AlternativeType Name="boost::unordered::concurrent_flat_set&lt;*&gt;" />
+        <AlternativeType Name="boost::unordered::concurrent_node_map&lt;*&gt;" />
```
k3DW, line 424: Another copy of this line needs to be inserted to 438 below, otherwise the `concurrent_node_map` will be displayed like a set. (i.e. using the indices in the item name as opposed to using the keys)

joaquintides: Thank you! Fixed.

include/boost/unordered/detail/foa/core.hpp
joaquintides: It is with this type of constructor:

```cpp
  table_core(
    std::size_t n=default_bucket_count,const Hash& h_=Hash(),
    const Pred& pred_=Pred(),const Allocator& al_=Allocator()):
    hash_base{empty_init,h_},pred_base{empty_init,pred_},
    allocator_base{empty_init,al_},arrays(new_arrays(n)),
    size_ctrl{initial_max_load(),0}
    {}
```
As it happens, `arrays` (in `arrays(new_arrays(n))`) is a non-static member function that uses the internal allocator, but _not_ other data members that have not been initialized yet (namely `size_ctrl`), but the compiler's analysis is not as fine-grained as that, seemingly. The situation would be similar to this example:

```cpp
struct X
{
  // calculate_b() called before X is fully initialized, but it's OK as b and c do not
  // form part of the calculation.
  
  X(int a_): a(a_), b(calculate_b()), c(0){}

  int calculate_b() { return 2*a; }
  
  int a;
  int b;
  int c;
};
```
What I find surprising is that this warning didn't trigger before, maybe it only happens for complex, non-inlined scenarios.

doc/unordered/unordered_node_map.adoc
```diff
@@ -537,6 +538,24 @@ and always calls `other.reset_stats()`.
 
 ---
 
+==== Move Constructor from concurrent_node_map
+
+```c++
+unordered_flat_map(concurrent_node_map<Key, T, Hash, Pred, Allocator>&& other);
```
pdimov, line 544: This should presumably say `unordered_node_map`?

include/boost/unordered/detail/foa/node_handle.hpp
```diff
@@ -153,7 +182,17 @@ struct node_handle_base
...
     }
 
-    allocator_type get_allocator()const noexcept{return al();}
+    allocator_type get_allocator()const
+    {
+#if defined(BOOST_GCC)
+      /* GCC lifetime analysis incorrectly warns about uninitialized
+       * allocator object under some circumstances.
+       */
+      if(empty())__builtin_unreachable();
```
pdimov, line 191: This seems fishy to me. Is calling `get_allocator` undefined behavior when `empty()`? Is the removal of the `noexcept` on purpose?

doc/unordered/unordered_node_map.adoc
joaquintides: Good catch, fixed.

include/boost/unordered/detail/foa/node_handle.hpp
joaquintides: * `get_allocator` on an empty node handle is UB, see [here](https://eel.is/c++draft/container.node#observers-11).
* I removed `noexcept` on purpose cause the standard does not have it (presumably because of the Lakos rule on narrow-contract functions). This removal is not materially connected to the `__builtin_unreachable` addition, really.
* The `__builtin_unreachable` addition is due to GCC incorrectly, and not systematically, warning about potentially uninitialized node handles [here](https://github.com/boostorg/unordered/blob/bd5eae3c1a97d5557e8a0b4bfc513d1eed3c14bd/test/unordered/node_handle_allocator_tests.cpp#L130) and [here](https://github.com/boostorg/unordered/blob/bd5eae3c1a97d5557e8a0b4bfc513d1eed3c14bd/test/unordered/node_handle_allocator_tests.cpp#L138): adding  `__builtin_unreachable` helped _but not always_, so I was forced to also add [this](https://github.com/boostorg/unordered/blob/bd5eae3c1a97d5557e8a0b4bfc513d1eed3c14bd/test/unordered/node_handle_allocator_tests.cpp#L7-L13). I don't see any problem with foolproofing `node_handle_base::get_allocator` as I've done, but if it looks smelly/ugly I can just remove it, of course --the GCC pragma in `node_handle_allocator_tests.cpp` has us covered.

k3DW:
I caused the merge conflict with my other PR, so @joaquintides gave me permission to resolve it here.

PR #277, opened 2024-09-01
Fixed `std::initializer_list` assignment issues for open-addressing containers

### LINKED ISSUE #276: Missing assignment operator=(std::initializer_list<value_type>)
The boost.unordered containers are missing the assignment operator taking an `std::initializer_list` argument. While it may seem redundant, this can be needed in a few special cases - in particular when using custom allocators which lack default initialization (as many stateful allocators are)

joaquintides:
Fixes #276.

### DIFF
--- a/include/boost/unordered/unordered_flat_map.hpp
+++ b/include/boost/unordered/unordered_flat_map.hpp
@@ -205,8 +205,8 @@ namespace boost {
 
       unordered_flat_map& operator=(std::initializer_list<value_type> il)
       {
-        clear();
-        insert(il);
+        this->clear();
+        this->insert(il.begin(), il.end());
         return *this;
       }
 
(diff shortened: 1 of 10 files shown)


include/boost/unordered/unordered_flat_map.hpp
```diff
@@ -202,6 +203,13 @@ namespace boost {
         return *this;
       }
 
+      unordered_flat_map& operator=(std::initializer_list<value_type> il)
+      {
+        clear();
+        insert(il);
```
k3DW, line 209: You could save a function call by instead writing
```cpp
insert(il.begin(), il.end());
```

pdimov:
That's fine in principle but I prefer writing tests that reflect actual use. Nobody's going to assign `std::initializer_list<value_type>()` to a container; more probable user code would be to assign something like `{{k1, v1}, {k2, v2}}`.

include/boost/unordered/unordered_flat_map.hpp
joaquintides: Done, thank you!

joaquintides:
> That's fine in principle but I prefer writing tests that reflect actual use. Nobody's going to assign std::initializer_list<value_type>() to a container; more probable user code would be to assign something like {{k1, v1}, {k2, v2}}.

Fixed [here](https://github.com/boostorg/unordered/pull/277/commits/cc8b40345ef78ea191c4e6976e0f8d0b4365131f).

PR #290, opened 2024-10-03
Use std::ref instead of copying the function parameter…

### LINKED ISSUE #285: `insert_(or|and)_[c]visit` for `std::initializer_list` copies visitation function objects
(Concurrent containers.) The following:

https://github.com/boostorg/unordered/blob/4c0aea983ebec4a9491381e7e4f3a9da844d19ba/include/boost/unordered/concurrent_flat_map.hpp#L482-L487

copies `f`, thus unnecessarily imposing copyability on `F`. By contrast, other `concurrent_flat_map` visitation-enabled operations don't have such requirement since they delegate on `concurrent_table` functions accepting the visitation function by forwarding reference:

```cpp
#include <boost/unordered/concurrent_flat_map.hpp>

struct NCF
{
  NCF() = default;
  NCF(const NCF&) = delete;
  NCF(NCF&&) = default;

  template<typename T>
  void operator()(const T&) const {}
};

int main()
{
  boost::concurrent_flat_map<int, int> m;

  // ok
  m.insert_or_visit({0, 0}, NCF{}); 

  // 'NCF::NCF(const NCF &)': attempting to reference a deleted function
  m.insert_or_visit({{0, 0}, {1, 0}}, NCF{});
}
```

This can be fixed by wrapping `f` into a `std::reference_wrapper`. Documentation should also reflect the change. This should wait for #283 to be merged because `*_and_[c]visit` operations are also affected (on purpose, in anticipation of this issue publication). Applies to all concurrent containers.

k3DW:
…in std::initializer_list overloads of insert_{or|and}_[c]visit

Closes #285

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,20 +57,23 @@ jobs:
           - { compiler: gcc-9,     cxxstd: '11,14,17',    os: 'ubuntu-22.04', install: 'g++-9' }
           - { compiler: gcc-10,    cxxstd: '11,14,17,20', os: 'ubuntu-22.04', install: 'g++-10' }
           - { compiler: gcc-11,    cxxstd: '11,14,17,20', os: 'ubuntu-22.04', install: 'g++-11' }
-          - { name: "gcc-12 w/ sanitizers (11)", sanitize: yes,
-              compiler: gcc-12,    cxxstd: '11',          os: 'ubuntu-22.04', ccache_key: "san1" }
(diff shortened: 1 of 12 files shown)


joaquintides (approved):

k3DW:
I've been looking into these test failures, but I can't determine what's going on here. It seems like the failures aren't related to my change, but I'm not certain. I won't merge this myself. I will let @joaquintides do this, to indicate there are no further problems with this PR

Issue #301, opened 2025-01-20
Incorrect documentation for erase_if namespace (or missing forward/alias declaration)


hirokiht:
Looking at documentation, one might think erase_if works in the namespace directly, i.e. `boost::erase_if(...)`, however the current implementation will only compile inside unordered namespace currently `boost::unordered::erase_if(...)`

joaquintides:
Solved in #303.

PR #303, opened 2025-01-22
added header synopses to the reference

joaquintides:
Added header synopses to the reference in the style of the standard. Solves #301.


doc/modules/ROOT/pages/ref.adoc
```diff
@@ -1,17 +1,39 @@
 [#reference]
 = Reference
 
-* xref:reference/unordered_map.adoc[unordered_map]
-* xref:reference/unordered_multimap.adoc[unordered_multimap]
-* xref:reference/unordered_set.adoc[unordered_set]
-* xref:reference/unordered_multiset.adoc[unordered_multiset]
-* xref:reference/hash_traits.adoc[hash_traits]
+* xref:reference/header_unordered_map_fwd.adoc[`<boost/unordered/unordered_map_fwd.hpp>` Synopsis]
```
cmazakas, line 4: Try putting the code backticks on the outside of the `xref` and tell me if you like how it looks.

image

In this image, the top two links are the updated version. The third one is the current look.

joaquintides: I can solve this in a more satisfactory way, for that I'd need for you to add this definition to the source `site.css`:

```
.inlink-code {
    font-family: Roboto Mono,monospace;
    font-size: 95%;
    color: inherit;
}
```

Can you do that?

cmazakas: Of course. You can do that locally yourself even.

All you gotta clone is the unordered-ui-bundle repo and then invoke `gulp bundle` in it. There's docs here: https://gitlab.com/antora/antora-ui-default#user-content-development-quickstart

joaquintides: I managed to do it without chaning the CSS file: 191adf4a3028b88010e9c4e735380fc79bddaffe.

joaquintides:
> The reference is awfully cramped here on my screen.
> 
> I did a width adjustment in the CSS and this is how it looks now: 
> [...]
> Let me know what you think. It's kind of nice how much we can really fine tune here.

My concern with that is that less space is reserved for the actual content, and the whole page's width is constrained in boost.io. Is it possible to have the TOC's width be adjustable by the user by dragging its right border?

cmazakas:
> > The reference is awfully cramped here on my screen.
> > I did a width adjustment in the CSS and this is how it looks now:
> > [...]
> > Let me know what you think. It's kind of nice how much we can really fine tune here.
> 
> My concern with that is that less space is reserved for the actual content, and the whole page's width is constrained in boost.io. Is it possible to have the TOC's width be adjustable by the user by dragging its right border?

Ah, that's a fair point. It looked okay when I sit the `width: auto;' but we can look into making it manually adjustable, yeah.

joaquintides:
> Ah, that's a fair point. It looked okay when I sit the `width: auto;' but we can look into making it manually adjustable, yeah.

As per our private conversation, moving this to another issue and merging this PR now.

PR #316, opened 2025-08-25
Update Link to regression test matrix in README

Flamefire:


joaquintides:
Hi, once https://github.com/boostorg/website-v2/issues/1901 is merged, this fix is not strictly required (links now work), but the links you provide are cleaner. Would you mind replacing `regression.boost.io` with `regression.boost.org` before I merge? Thank you!

Flamefire:
> Would you mind replacing `regression.boost.io` with `regression.boost.org`

Can do but what is the reason for that? From the linked issue:
> configmap-nginx.yaml: location = /development/tests/master/developer/issues.html { return 301 https://regression.boost.io/master/developer/issues.html; }

So it does use boost.io there too which is why I used it as the "canonical" URL

joaquintides:
If I click on 

https://regression.boost.org/master/developer/issues.html

the browser displays "boost.org" rather than "boost.io":

imagen

So, the boost.io domain is effectively unknown to the user. I understand boost.io will be removed in the future as everything should go under boost.org as it always did.

Flamefire:
Ouch, my bad. Thanks for the explanation

PR #317, opened 2025-10-05
Mark natvis file as PUBLIC in CMakeLists.txt

### LINKED ISSUE #307: Natvis not picked up on vcpkg
The current cmake script seems to be failing to pass the natvis to VS when consumed by vcpkg.
I've had to manually copy the natvis file to my source directory to get it loaded.

I'm guessing this is due to unordered being an interface in cmake while also marking the natvis file private, so there is neither a lib to consume, nor a config to pass down to consuming projects.

Unfortunately, without that natvis file, unordered_* is not a pleasant experience while debugging :/

(Version: 1.87.0, did not work on 1.86.0 either)

k3DW:
Closes #307 , in tandem with boostorg/cmake#89


CMakeLists.txt
```diff
@@ -23,13 +23,17 @@ target_link_libraries(boost_unordered
     Boost::throw_exception
 )
 
-if(CMAKE_VERSION VERSION_GREATER 3.18 AND CMAKE_GENERATOR MATCHES "Visual Studio")
-
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.12)
```
pdimov: You don't need this check.

CMakeLists.txt
```diff
@@ -23,13 +23,17 @@ target_link_libraries(boost_unordered
...
   file(GLOB_RECURSE boost_unordered_IDEFILES CONFIGURE_DEPENDS include/*.hpp)
+else()
+  file(GLOB_RECURSE boost_unordered_IDEFILES include/*.hpp)
+endif()
+target_sources(boost_unordered PRIVATE ${boost_unordered_IDEFILES})
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
```
pdimov: This is wrong. `target_sources` requires 3.19, not `source_group` and not 3.18.

CMakeLists.txt
pdimov: I mean, CONFIGURE_DEPENDS "works" on pre-3.12, because it's interpreted as a path, and doesn't do anything.

PR #332, opened 2025-12-31
Docs: download extension

sdarwin:
When building antora docs, retry the UI download if it fails.

Generally, any network downloads in any build scripts should be configured to retry.   

For example, `curl` has options `--retry` and `--retry-all-errors`.   

source code: https://github.com/cppalliance/antora-downloads-extension

### DIFF
--- a/doc/package-lock.json
+++ b/doc/package-lock.json
@@ -5,6 +5,9 @@
     "packages": {
         "": {
             "name": "doc",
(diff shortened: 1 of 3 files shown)


joaquintides:
Could you plase update `package-lock.json` before I merge? See https://github.com/boostorg/unordered/pull/335

PR #334, opened 2026-01-01
Fix issue with constructing an FOA container from a class with a template conversion operator

### LINKED ISSUE #331: unordered_node_map.hpp compilation failure with clang due to concurrent_node_map_fwd.hpp include
When returning an `unordered_node_map` in a `boost::leaf::result` from a function, clang won't compile due to incomplete type related to concurrent_node_map.

Also applies to `unordered_flat_map`.

Please let me know if this should be filed in the leaf repo instead.

Building with clang 22 using `--gcc-toolchain=<path-to-gcc-15.2.0>` 

Reproducible Example
```cpp
#include "boost/unordered/unordered_node_map.hpp"
#include "boost/leaf/result.hpp"

template <typename T>
boost::leaf::result<T> make_result_error()
{
  return boost::leaf::new_error();
}

int main()
{
  using map_t = boost::unordered::unordered_node_map<std::string, std::string>;

  auto result = make_result_error<map_t>();

  return 0;
}
```

Error
```
auto result = make_result_error<map_t>();
.../boost_1_90_0/include/boost/unordered/concurrent_node_map_fwd.hpp:32:11: note: template is declared here
   class concurrent_node_map;
In file included from .../boost_1_90_0/include/boost/unordered/unordered_node_map.hpp:14:
In file included from .../boost_1_90_0/include/boost/unordered/concurrent_node_map_fwd.hpp:19:
In file included from .../gcc-15.2.0/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/functional:51:
In file included from .../gcc-15.2.0/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/bits/stl_function.h:60:
In file included from .../gcc-15.2.0/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/bits/move.h:37:
.../gcc-15.2.0/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/type_traits:1159:21: error: static assertion failed due to requirement 'std::__is_complete_or_unbounded(std::__type_identity<boost::unordered::concurrent_node_map<std::basic_string<char, std::char_traits<char>, std::allocator<char>>, std::basic_string<char, std::char_traits<char>, std::allocator<char>>, boost::hash<std::basic_string<char, std::char_traits<char>, std::allocator<char>>>, std::equal_to<std::basic_string<char, std::char_traits<char>, std::allocator<char>>>, std::allocator<std::pair<const std::basic_string<char, std::char_traits<char>, std::allocator<char>>, std::basic_string<char, std::char_traits<char>, std::allocator<char>>>>>>{})': template argument must be a complete class or an unbounded array
static_assert(std::__is_complete_or_unbounded(__type_identity<_Tp>{}),
```

k3DW:
Closes #331

### DIFF
--- a/include/boost/unordered/concurrent_flat_map.hpp
+++ b/include/boost/unordered/concurrent_flat_map.hpp
@@ -2,6 +2,7 @@
  *
  * Copyright 2023 Christian Mazakas.
  * Copyright 2023-2026 Joaquin M Lopez Munoz.
+ * Copyright 2026 Braden Ganetsky
  * Distributed under the Boost Software License, Version 1.0.
  * (See accompanying file LICENSE_1_0.txt or copy at
  * http://www.boost.org/LICENSE_1_0.txt)
@@ -189,10 +190,10 @@ namespace boost {
       {
       }
 
-
-      template <bool avoid_explicit_instantiation = true>
+      template <typename Key2,
+        typename std::enable_if<std::is_same<Key, Key2>::value, int>::type = 0>
       concurrent_flat_map(
-        unordered_flat_map<Key, T, Hash, Pred, Allocator>&& other)
+        unordered_flat_map<Key2, T, Hash, Pred, Allocator>&& other)
           : table_(std::move(other.table_))
       {
       }
--- a/include/boost/unordered/concurrent_flat_set.hpp
+++ b/include/boost/unordered/concurrent_flat_set.hpp
@@ -2,6 +2,7 @@
  *
  * Copyright 2023 Christian Mazakas.
  * Copyright 2023-2026 Joaquin M Lopez Munoz.
+ * Copyright 2026 Braden Ganetsky
  * Distributed under the Boost Software License, Version 1.0.
  * (See accompanying file LICENSE_1_0.txt or copy at
  * http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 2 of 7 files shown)


joaquintides:
Not entirely sure, but wouldn't a workaround based on `boost::unordered::detail::type_identity` achieve the same result with less hassle?

k3DW:
> @k3DW it doesn't look like we're going to come up with a totally transparent solution right now, and Boost 1.91 is approaching, so, given that the reported issues looks important enough, I'd go with something like [#334 (comment)](https://github.com/boostorg/unordered/pull/334#issuecomment-3704121330) and maybe file an issue to potentially continue investigating after the release.

I wasn't able to get to this in time for 1.91. I've updated this PR. I'll open a new issue. Even if we don't fix it soon, it'll be documented that this is a limitation.

k3DW:
I'm not fully convinced that we need an issue to follow up. I don't know whether it matters. This change, as I have it right now, breaks certain constructors through conversion operators.

```cpp
using flat_map = boost::unordered::unordered_flat_map<int, int>;
using c_flat_map = boost::unordered::concurrent_flat_map<int, int>;

struct converter
{
  operator c_flat_map();
};

static_assert(std::is_constructible_v<flat_map, converter>));
```

The use case above worked before this PR, and will be broken after it. But should we even try to support esoteric use cases like this? It's better to be more explicit in this case of a double conversion, from `converter` to `c_flat_map`, then to `flat_map`.

Note that constructing `flat_map` from a class that has a conversion operator to `flat_map` would still work.

joaquintides:
> I'm not fully convinced that we need an issue to follow up. I don't know whether it matters.

I agree with you. Merging now, let's see if we can get permission to get it into Boost 1.91. Thank you!

PR #336, opened 2026-01-02
simplified doc/Jamfile.v2

### LINKED ISSUE #330: duplicate documentation
Download and unzip a copy of boost 1.90.0 from https://archives.boost.io/release/1.90.0/source/

```
cd boost_1_90_0/libs/unordered/doc/html
ls -al
cd boost_1_90_0/libs/unordered/doc/html/unordered
ls -al
```
The same files exist in both doc/html and doc/html/unordered.   
Where should they go...  

It seems like doc/html would be enough, and the last extra "unordered/" is superfluous.  Just put documentation in "doc/html" (unless there are other reasons. maybe there are.)  

A boost library is an example for other libraries. `boostorg/openmethod` now has the same issue. both should be fixed.

joaquintides:
* Removed `install` step as it duplicated generated HTML files and Antora already puts its output in the final destination anyway.
* Deletes subdirectory `node_modules` after Antora generation: this directory has slipped previously in release archives (seemingly, only `bin` directories are automatically located and removed).

Fixes #330.

### DIFF
--- a/doc/Jamfile.v2
+++ b/doc/Jamfile.v2
@@ -1,57 +1,23 @@
-import generate ;
-import path ;
-import property-set ;
-import virtual-target ;
+# Copyright 2026 Joaquin M Lopez Munoz
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-path-constant HERE : . ;
+make html_ : build_antora.sh : @run-script ;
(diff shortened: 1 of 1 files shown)


sdarwin:
> Backwards compatibility

OK.  

I think, in the future, if the "redirect to latest version" implements more logic to check the presence or absence of the redirect file, it could also include logic to fix `unordered` as an exception.  However a prerequisite is getting a basic feature working,  which is not there yet.

PR #355, opened 2026-06-11
added interoperability with C++20 ranges

### LINKED ISSUE #347: C++23 insert_range and std::from_range_t constructors
`unordered` containers currently have no C++23 insert_range method or constructors with std::from_range_t. These features don't look like something completely incompatible with these containers, so maybe they can be added?

joaquintides:
Added interoperability with C++20 ranges to all the containers in the library: `insert_range` (plus `insert_range_(or|and)_[c]visit` for concurrent containers), `std::from_range` construction and associated CTAD deduction guides. The `boost::unordered::from_range` construction tag is provided as an alternative to C++23 `std::from_range` or when this is not available.

Closes #347.

### DIFF
--- a/doc/modules/ROOT/pages/reference/concurrent_flat_map.adoc
+++ b/doc/modules/ROOT/pages/reference/concurrent_flat_map.adoc
@@ -77,6 +77,12 @@ namespace unordered {
     xref:#concurrent_flat_map_copy_constructor_with_allocator[concurrent_flat_map](const concurrent_flat_map& other, const Allocator& a);
     xref:#concurrent_flat_map_move_constructor_with_allocator[concurrent_flat_map](concurrent_flat_map&& other, const Allocator& a);
     xref:#concurrent_flat_map_move_constructor_from_unordered_flat_map[concurrent_flat_map](unordered_flat_map<Key, T, Hash, Pred, Allocator>&& other);
+    template<class FromRangeT, xref:#concurrent_flat_map_container_compatible_range[__container-compatible-range__]<value_type> R>
+      xref:#concurrent_flat_map_range_constructor[concurrent_flat_map](FromRangeT&&, R&& rg,
+                          size_type n = _implementation-defined_,
+                          const hasher& hf = hasher(),
+                          const key_equal& eql = key_equal(),
+                          const allocator_type& a = allocator_type());
     xref:#concurrent_flat_map_initializer_list_constructor[concurrent_flat_map](std::initializer_list<value_type> il,
                         size_type n = _implementation-defined_
                         const hasher& hf = hasher(),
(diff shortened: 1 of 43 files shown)


doc/modules/ROOT/pages/reference/concurrent_flat_map.adoc
```diff
@@ -591,6 +655,27 @@ Complexity:;; O(`bucket_count()`)
...
+  concurrent_flat_map(FromRangeT&&, R&& rg,
+                      size_type n = _implementation-defined_,
+                      const hasher& hf = hasher(),
+                      const key_equal& eql = key_equal(),
+                      const allocator_type& a = allocator_type());
+----
+
+Constructs an empty table with at least `n` buckets, using `hf` as the hash function, `eql` as the key equality predicate and `a` as the allocator, and inserts the elements from `rg` into it.
+
+Only participates in overload resolution if `FromRangeT` is convertible to `std::from_range_t` or `boost::concurrent::from_range_t`.
```
pdimov, line 671: Why convertible to?

joaquintides: Call me anal, but to preserve this:
```cpp
#include <vector>

struct foo: std::from_range_t { foo() = default; };

int main()
{
  std::vector<int> x;
  std::vector<int> y(foo{}, x);
}
```

include/boost/unordered/detail/ranges_support.hpp
```diff
@@ -0,0 +1,101 @@
...
+namespace unordered {
+
+struct from_range_t { explicit from_range_t() = default; };
+inline constexpr from_range_t from_range{};
+
+namespace detail {
+
+/* Code below picks std::from_range_t if it exists,
+ * boost::unordered::from_range_t otherwise. Technique explained at
+ https://bannalia.blogspot.com/2016/09/compile-time-checking-existence-of.html
```
pdimov, line 46: That's interesting but do we need it? Isn't a feature macro good enough for detecting whether `std::from_range_t` exists?

joaquintides: I don't think we have such macro? Please correct me if I'm wrong.

joaquintides: Umm, looks like `__cpp_lib_containers_ranges` is the thing, I can try using that instead of the crazy workaround.

doc/modules/ROOT/pages/reference/concurrent_flat_map.adoc
pdimov: I know that's the reason, but I'm wondering why do we want to preserve that.

joaquintides: Well, maximum conformance to the std spec. I can use a simpler concept checking for exact equality, but it's the same amount of code.

pdimov: OK then. It's probably not worth deviating.

include/boost/unordered/detail/ranges_support.hpp
joaquintides: Ok, no dice:

https://godbolt.org/z/8asY1doo7
```cpp
// GCC 14.1
#include <ranges>

#if !defined(__cpp_lib_containers_ranges)
#pragma message ("__cpp_lib_containers_ranges not defined")
#endif

int main()
{
  (void)std::from_range;
}
```
```
<source>:4:59: note: '#pragma message: __cpp_lib_containers_ranges not defined'
    4 | #pragma message ("__cpp_lib_containers_ranges not defined")
      |                                                           ^

Program returned: 0
```

pdimov: Eh. Well, then add a comment explaining why we're using the crazy workaround instead of the feature macro. :-)

joaquintides: Resolved by 610879b1270f792193b5b99f469e6151feff8eb0.

